PR #5, opened 2017-02-01
Remove erroneous typeof value === 'integer'

sandersn:
in src/services/jsonCompletion:JSONCompletion.getInsertTextForGuessedValue

This will be an error in an upcoming version of Typescript (2.2 or 2.3 probably).


aeschli:
Thanks @sandersn !

PR #7, opened 2017-06-17
Add null check to prevent service from crashing

adamvoss:
`Strings.startsWith` throws on null values.  As a result, JSON Language Server will crash if `$schema` is set to `null`.  This change adds appropriate checking so the service fails gracefully.


aeschli:
Thanks a lot, @adamvoss !

PR #13, opened 2018-01-09
Correct grammar to 'String is not a URI'

stepjanssen:
Old message was 'String is not an URI'. Unless 'URI' should be pronounced differently from what I'm used to, it should be 'a URI'.


aeschli:
Thanks @stepjanssen !

PR #14, opened 2018-02-07
fix errors when jsonDocument.syntaxErrors is not defined

kkhcheng:
I found this error in vscode when opening yaml files


aeschli:
Fixed by a different change, but thanks for the PR!

PR #17, opened 2018-03-20
Fix issue related to jsonDocument.syntaxErrors

nervebassmaster:
Got `The YAML Language Server server crashed 5 times in the last 3 minutes. The server will not be restarted.` when started VSCode.

That issue is related to the following error: `.vscode/extensions/adamvoss.yaml-0.0.10/server/node_modules/vscode-json-languageservice/lib/services/jsonValidation.js:47 jsonDocument.syntaxErrors.forEach(addProblem); `
`TypeError: Cannot read property 'forEach' of undefined....`

Added expression to check whether `jsonDocument.syntaxErrors` is defined.

Ran tests. Pass as well.
Linted code. No issues found.

### DIFF
--- a/src/services/jsonValidation.ts
+++ b/src/services/jsonValidation.ts
@@ -83,12 +83,14 @@ export class JSONValidation {
 				}
 			}
 
-			jsonDocument.syntaxErrors.forEach(p => {
-				if (p.code === ErrorCode.TrailingComma) {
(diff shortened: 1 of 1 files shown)


aeschli:
@nervebassmaster The contract is that JSONDocument.syntaxErrors is never null as established in the constrctor of JSONDocument.
That should be fixed in the yaml server.

PR #18, opened 2018-03-21
fix Cannot read property 'forEach' of undefined

dshvedchenko:
to avoid crash https://github.com/adamvoss/vscode-yaml/issues/10

### DIFF
--- a/src/services/jsonValidation.ts
+++ b/src/services/jsonValidation.ts
@@ -82,13 +82,14 @@ export class JSONValidation {
 					trailingCommaSeverity = commentSeverity = ProblemSeverity.Ignore;
 				}
 			}
-
-			jsonDocument.syntaxErrors.forEach(p => {
-				if (p.code === ErrorCode.TrailingComma) {
-					p.severity = trailingCommaSeverity;
-				}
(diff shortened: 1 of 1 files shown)


aeschli:
duplicate of https://github.com/Microsoft/vscode-json-languageservice/pull/17

This should be fixed in the Yaml plugin. They implement internal API that has changed.

dfee:
@aeschli I don't have "YAML Support by RedHat" installed. Is that the plugin you're suggesting should be fixed? Otherwise, can you link to what other repository needs a change so folks can contribute?

image

aeschli:
The problem is here: https://github.com/adamvoss/vscode-yaml-languageservice/blob/master/src/parser/yamlParser.ts
This code implements the internal JSONDocument that has changed in the meantime. 

The easiest fix would be depend on vscode-json-languageservice@2.0.17 which is the last version of the old signature.

I created https://github.com/adamvoss/vscode-yaml-languageservice/pull/9

aeschli:
New APIs have been created to enable the YAML language server to create a JSONDocument from a custom ASTNode.

PR #21, opened 2018-05-07
fix maxItems validation error message

sqs:
The message refers to the `minItems` value (incorrect), not the `maxItems` value (correct).

This leads to incorrect validation error messages when using maxItems, such as in the example below (which can occur with `maxItems` is not 0):

> Array has too many items. Expected 0 or fewer.


aeschli:
Thanks @sqs !

PR #24, opened 2018-08-13
[draft-07] Implement if/then/else

Fer0x:
The current version of JSON Schema is draft-07 and it has some awesome features like conditional schema evaluation.


aeschli:
Very cool, thanks @Fer0x !

PR #27, opened 2018-09-05
Use 'if' to create completions

Levertion:
I'm not certain that this implementation is correct for the codebase, but it doesn't break any of the other tests (as added in #24) and passes my new test. @fer0x: if you could validate that this looks right too.

I've only added one test case, but want to check that this feature is desired before adding more tests.

### DIFF
--- a/src/parser/jsonParser.ts
+++ b/src/parser/jsonParser.ts
@@ -469,11 +469,11 @@ function validate(node: ASTNode, schema: JSONSchema, validationResult: Validatio
 			testAlternatives(schema.oneOf, true);
 		}
 
-		let testBranch = (schema: JSONSchema) => {
+		let testBranch = (schema: JSONSchemaRef) => {
 			let subValidationResult = new ValidationResult();
 			let subMatchingSchemas = matchingSchemas.newSub();
 
-			validate(node, schema, subValidationResult, subMatchingSchemas);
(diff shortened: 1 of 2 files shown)


Fer0x:
@Levertion works correct for me. Thanks for enhancement.

aeschli:
@Levertion The test is still failing...

Levertion:
OK - I've written enough tests to convince me that my behaviour is correct, although I'm unsure about the formatting.

Should we format automatically somehow - clearly the style is to use `'`, but my personal habit is to use `"` so that is what I have used. I would expect there to be automatic formatting, but the builtin `vscode` formatting changes the style of the rest of the file, `prettier` is not being used and the `tslint` rules used does not control formatting.

My personal preference is [`prettier`](https://prettier.io/), but if the current formatting needs manually editing, let me know.

aeschli:
just use the built-in TypeScript formatter with default settings

Levertion:
I have tried that, using both format document and format selection after commenting out my entire `settings.json` and disabling all extensions for the workspace. This however led to only the minimal diff seen in 9f3ac837755d0a8b5aacd0552164b4932a246333, which doesn't seem optimal. Is that enough?

aeschli:
Yes, that's fine. The formatter preserves new lines, that's why the changes are so small.

aeschli:
It's great, thanks a lot @Levertion !

PR #30, opened 2018-10-24
Add a more sensible fallback range if there is no parent property

Levertion:
EDIT: After some consideration I realise this should be an issue rather than a pull request, as this is not obviously better in all cases.

Before:
```
{"bold":true}
^
Missing property "text"
```
After:
```
{"bold":true}
^^^^^^^^^^^^^
Missing property "text".

This case could occur if the entire document was validated against a schema using `required` or `oneOf`. This would have the disadvantage of making the errors uglier in cases where there is a large document which fails this validation, however only erroring the first character is also not great. Perhaps we should work out an alternative method?

### DIFF
--- a/src/parser/jsonParser.ts
+++ b/src/parser/jsonParser.ts
@@ -449,7 +449,7 @@ function validate(node: ASTNode, schema: JSONSchema, validationResult: Validatio
 
 			if (matches.length > 1 && maxOneMatch) {
 				validationResult.problems.push({
-					location: { offset: node.offset, length: 1 },
+					location: { offset: node.offset, length: node.length },
 					severity: DiagnosticSeverity.Warning,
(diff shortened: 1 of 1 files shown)


aeschli:
Multi line squiggles or even the whole document being red is really ugly and that was the reason to make the error that small.
I'm not really in flavor of this change.

Levertion:
Yeah, that's entirely reasonable. My use-case is for another format with inline json support on one line, where the single space squiggle appears to be too small, but it becomes ugly for a larger string anyway.

There is no good solution, but I'll close this.

Levertion:
Would it be worth having an heurstic such as if the node length is less than, say 35 we give an error over the whole range, else give an error on the starting character - for smallish sections the small error range is somewhat unfortunate.

PR #32, opened 2018-10-24
More concise error message for schema validation errors

xxyy:
#### Ellipsis
The error message shown if a schema could not be retrieved is currently not very concise, it repeats the same sentence three times. This PR shortens it a little.

ref: https://github.com/Microsoft/vscode/pull/60219#issuecomment-428089447 by @aeschli 

Old message:
bad

New message:
image

How this works is that it strips everything before `Error: ` in the message returned by the backend, if `Error: ` appears in the message. In the messages I observed, the removed part was something like `Unable to connect to <url>: Error: `, which is redundant because it's prepended by the [only caller](https://github.com/Microsoft/vscode-json-languageservice/blob/master/src/services/jsonSchemaService.ts#L404-L405) anyways.

I wasn't sure about the exact desired format of the error message, so if another format is better, I can change it. One issue with the current one is that it doesn't explicitly state that a schema is the problem. However I didn't want to change an already-translated message since translations would need to update.

### DIFF
--- a/src/services/jsonSchemaService.ts
+++ b/src/services/jsonSchemaService.ts
@@ -354,7 +354,12 @@ export class JSONSchemaService implements IJSONSchemaService {
 				return new UnresolvedSchema(schemaContent, errors);
 			},
 			(error: any) => {
-				let errorMessage = localize('json.schema.unabletoload', 'Unable to load schema from \'{0}\': {1}', toDisplayString(url), error.toString());
+				let errorMessage = error.toString();
+				let errorSplit = error.toString().split('Error: ');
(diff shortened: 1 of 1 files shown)


aeschli:
Change makes sense. Thanks!

xxyy:
Apparently, there was a test for the exact message, which fails now: https://travis-ci.org/Microsoft/vscode-json-languageservice/builds/445672376?utm_source=github_status&utm_medium=notification

I didn't see that there were tests I needed to run, the correct build command (I used `yarn run compile`) would be a good addition to the project readme.

I'll create another one to fix that real quick.

PR #34, opened 2018-10-24
Add an editorconfig

BackEndTea:
This editorconfig reflects the current style of the project.

See also: 
* https://editorconfig.org/
* https://backendtea.github.io/add-an-editorconfig

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,21 @@
+root = true
+
+[*]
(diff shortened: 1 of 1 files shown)


aeschli:
I'd rather not take that PR as the main contributers don't editor config it and it would get out-of date.
We're using the vscode default settings.

PR #35, opened 2018-11-08
Fixed a bug when comments are allowed but trailing commas aren't

abc-55:
This was probably a typo. This change doesn't affect VSCode because in VSCode, either both comments and trailing commas are allowed, or both are not allowed.

### DIFF
--- a/src/services/jsonValidation.ts
+++ b/src/services/jsonValidation.ts
@@ -77,7 +77,7 @@ export class JSONValidation {
 
 			jsonDocument.syntaxErrors.forEach(p => {
 				if (p.code === ErrorCode.TrailingComma) {
-					if (typeof commentSeverity !== 'number') {
+					if (typeof trailingCommaSeverity !== 'number') {
 						return;
 					}
 					p.severity = trailingCommaSeverity;


abc-55:
I have just discovered that this would change the behavior of Monaco:

- Monaco sets `allowComments` to `true` (https://github.com/Microsoft/monaco-json/blob/6653d96ddd9d5659147b5a096daf292685d1e66f/src/monaco.contribution.ts#L46)
- Monaco doesn't provide `documentSettings` (https://github.com/Microsoft/monaco-json/blob/39e2f4e7093adebbd6ba5861aa524c97d38012c5/src/jsonWorker.ts#L68)

This means that `trailingCommaSeverity` is `DiagnosticSeverity.Error` and `commentSeverity` is `undefined`
https://github.com/Microsoft/vscode-json-languageservice/blob/7b0bc0dcf45ab6411429e986716f969f92773850/src/services/jsonValidation.ts#L52-L53

However, because of the bug, trailing commas were ignored too.

The question is: What is the right behavior? Should `allowComments` also affect trailing commas?

PR #43, opened 2019-10-04
Add support for loading schemas stored as YAML files

hreeder:
Originally raised against [redhat-developer/yaml-language-server](https://github.com/redhat-developer/yaml-language-server).
This adds support for loading schemas saved in YAML formatted files, while still preserving the preference for JSON to be loaded first.

The logic here is to attempt to load as JSON first. If loading as JSON fails, attempt loading as YAML. If YAML loading succeeds, clear the list of errors. If YAML loading fails then add the YAML error to the errors array.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -61,7 +61,6 @@
       "version": "1.0.10",
       "resolved": "https://registry.npmjs.org/argparse/-/argparse-1.0.10.tgz",
       "integrity": "sha512-o5Roy6tNG4SL/FOkCAN6RzjiakZS25RLYFrcMttJqbdd8BWrnA+fGz57iN5Pb06pvBGvl5gQ0B48dJlslXvoTg==",
-      "dev": true,
       "requires": {
         "sprintf-js": "~1.0.2"
       }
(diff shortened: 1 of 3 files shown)


aeschli:
We have the SchemaRequestService to fetch schemas. We could change it so that it can return a string but also an object. That way the code to parse YAML stays on your side. I'd rather not ship a YAML parser with the jsonl-langaugeservice. 

What do you think?

hreeder:
That makes sense - I did feel that adding an entire YAML parser dependency was sort of over the top for a single feature.

As for actual implementation of what you're suggesting - I'll have to admit to being quite new to the codebase so I don't entirely understand. Would you essentially be suggesting doing the parsing within the yaml-language-server project instead?

The only usecase I can think of (off the top of my head) for adding it wider is if someone was writing JSON files but their schema stored in YAML (though I suspect this is a very small set of users)

aeschli:
Yes, then the parsing happens in the yaml-language-server, as part of the [SchemaRequestService](https://github.com/microsoft/vscode-json-languageservice/blob/2d9fffe5b008d315400bc3a51472abc794e51d8b/src/jsonLanguageTypes.ts#L137)
So far I haven't come across schemas written in YAML, but if that is a common use case also for JSON files, then we can reconsider (or have the json-languageservice have a yaml parser as well).

Let me know you want to adopt the PR, or you want me to make the change.

hreeder:
Happy to take this on myself.

Looking through old issues there is redhat-developer/vscode-yaml#182 - but I don't know if it's a common enough usecase to upstream in that case.

I'll close this PR now and work on an updated implementation against the YAML LS. Thanks for your guidance @aeschli, much appreciated!

PR #47, opened 2019-12-18
Expose getMatchingSchemas

ianchi:
Expose a new method `getMatchingSchemas` that resolves the applicable schema for the document and return the matching schema for each document node.

This is very useful for extending the functionality of the validator and adding custom keys.

I need to do some custom validation and exposing this simple function would make it super easy to implement.
Otherwise I'd need to duplicate a lot of code.


ianchi:
Fixed error when no schema matches.
Additionally export some members for `jsonDocument`.
They are also needed to easily extend functionality.

aeschli:
@ianchi I'm very hesitant to make `getMatchingSchemas` API in the current form. 
Can you provide some use cases?

ianchi:
Hi @aeschi,

I have a schema that has a custom keyword that triggers a custom validation logic.
I need to do this validation on the input files and report errors on a cli tool and also I made an extension that adds the errors as diagnostics to VSCode.

Having the matching schemas makes all this dead simple.
Without that I would have to rewrite all the schema validation logic, as there is no extension mechanism.

This can be validated with custom keywords using Ajv, but it is really not useful for reporting document position of the error.

I understand that it exposes some implementation details, but it allows to use it to easily extend functionality.
Alternatively some hook could be implemented to call custom functions, but I understood from some other Issue, that you weren’t willing to do so.

Exposing some of jsonDocument functionality also allows me to easily add some custom hover info with the vscode extension, that is dependant on the schema of the cursor’s node

ianchi:
Hi @aeschli, any chance of considering this?

Or any other approach to leverage this service and be able to extend it to use it on custom validations an hoover info?

aeschli:
Hi @ianchi , I pushed a commit that hides the inverted schema. Also I polished the code and added tests.

If you can sign the CLA, then I can push the change.

ianchi:
Hi @aeschli,
Just signed the CLA.
Great to see this merged.

Thanks!

PR #50, opened 2020-03-07
add support for textDocument/definition

ananthakumaran:
JSON Schema, Swagger, etc use $ref to reference reusable json snippets. Add the ability to jump to definition from $ref.

ezgif com-video-to-gif


ananthakumaran:
updated

aeschli:
Thanks @ananthakumaran !

PR #65, opened 2020-07-29
Use propertyNames for completion proposals

KapitanOczywisty:
This update will add completion for names provided in propertyNames' `enum` and `const` fields, with respect of `doNotSuggest` and `deprecationMessage`. Appropriate tests included.


aeschli:
@KapitanOczywisty Great stuff, thanks!

PR #67, opened 2020-08-19
Fix multipleOf with float

KapitanOczywisty:
This update improves how floats are handled with multipleOf option, based on @aeschli concept (linked issue).

One more optional change is to replace `val % multipleOf === 0` to `Number.isInteger(val / multipleOf)`, which allows a bit bigger numbers. Edge case example: 
`6000000000300000000 % 20000000001 = 256 // wrong` vs `6000000000300000000 / 20000000001 = 300000000 // fine`

Fix microsoft/vscode#72314


aeschli:
Really cool, thanks a lot!

PR #69, opened 2020-08-21
Resolve refs inside additionalItems

KapitanOczywisty:
`additionalItems` was not checked for `$ref`s.

fix microsoft/vscode#69071 fix microsoft/vscode#75394

### DIFF
--- a/src/services/jsonSchemaService.ts
+++ b/src/services/jsonSchemaService.ts
@@ -509,7 +509,7 @@ export class JSONSchemaService implements IJSONSchemaService {
 					}
 				}
 
-				collectEntries(<JSONSchema>next.items, <JSONSchema>next.additionalProperties, next.not, next.contains, next.propertyNames, next.if, next.then, next.else);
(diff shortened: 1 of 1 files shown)


aeschli:
Oh wow, cood catch!

unional:
Just noticed this and want to point out that:

> `additionalItems` doesn’t make sense if you’re doing “list validation” (items is an object), and is ignored in the case.

aeschli:
@unional When looking for `$ref`s to resolve, we always visit all properties, regardless if they are applicable. That might be unnecessary work, but I don't think is a problem. Let me know if you think otherwise.

unional:
Yeah, I notice that is the design of the code. The only concern is about "soundness", as language-service is used to provide runtime validation, so people writing json schema and rely on IDE to provide feedback will miss this case until they run other validation that catches this.

KapitanOczywisty:
@unional This extension provides intellisense and somewhat accurate validation, but if you want to be sure that it'll run fine on other validation tools, you need to write some tests. Here there are implemented features from different drafts and even non-standard stuff added for vscode use (error messages). I don't think that anyone is willing to bring it closer to drafts, since it's good enough for intellisense and there are very easy to use packages like ajv.

unional:
Sure, no problem. And regarding drafts, really look forward for 2020-12 support. 😆

PR #70, opened 2020-08-21
Resolve schemas from getMatchingSchemas argument

### LINKED ISSUE #68: getMatchingSchemas with unresolved schema
Great addition with `getMatchingSchemas`, but there is small inconsistency with how schema is handled i.e. when schema is provided as an argument, `$ref`s are not resolved. Example bellow.

So maybe we need to expose resolver api or call resolver inside getMatchingSchemas?

```js
const jsonSchema = {
  definitions: {
    field: {
      type: "number",
    },
  },
  type: "array",
  items: { $ref: "#/definitions/field" },
};
const jsonCode = "[1,2]";

ls.configure({
  validate: true,
  allowComments: true,
  schemas: [
    {
      uri: "test://schema.json",
      schema: jsonSchema,
      fileMatch: ["*.json"],
    },
  ],
});

const document = TextDocument.create("test://test.json", "json", 0, jsonCode);
const jsonDoc = ls.parseJSONDocument(document);

const show = (list) => {
  console.log(list.map((x) => JSON.stringify(x.schema)));
};

ls.getMatchingSchemas(document, jsonDoc, jsonSchema).then(show);
// unresolved $refs
// [
//   '{"$ref":"#/definitions/field"}',
//   '{"$ref":"#/definitions/field"}',
//   '{"definitions":{"field":{"type":"number"}},"type":"array","items":{"$ref":"#/definitions/field"}}'
// ]

ls.getMatchingSchemas(document, jsonDoc).then(show);
// resolved
// [
//   '{"type":"number"}',
//   '{"type":"number"}',
//   '{"definitions":{"field":{"type":"number"}},"type":"array","items":{"type":"number"}}'
// ]
```

KapitanOczywisty:
Schema from argument was not resolved. I've used similar code to `doValidation` method.

fix #68

### DIFF
--- a/src/services/jsonSchemaService.ts
+++ b/src/services/jsonSchemaService.ts
@@ -583,7 +583,10 @@ export class JSONSchemaService implements IJSONSchemaService {
 
 	public getMatchingSchemas(document: TextDocument, jsonDocument: Parser.JSONDocument, schema?: JSONSchema): Thenable<MatchingSchema[]> {
 		if (schema) {
-			return this.promise.resolve(jsonDocument.getMatchingSchemas(schema).filter(s => !s.inverted));
+			const id = schema.id || ('schemaservice://untitled/matchingSchemas/' + idCounter++);
+			return this.resolveSchemaContent(new UnresolvedSchema(schema), id, {}).then(resolvedSchema => {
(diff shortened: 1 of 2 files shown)


aeschli:
Thanks @KapitanOczywisty !

PR #72, opened 2020-08-27
fix silent errors on getMatchingSchemas

Trias:
Thanks for exposing getMatchingSchemas tot he public! it helped us solve our use case (expanding the json schema with custom attributes)

So this story happened: Seemingly randomly getMatchingSchemas returned only a subset of node we were expecting. What made matters worse, we failed to make a minimal reproduction, which made debugging very hard (we are applying the schema to roughly 10.000 files, the schema itself is split over several files).

Turns out, we had too many open files open at the same time, which blocked schema resolution. Unfortunately these errors were silenced, which made debugging very hard. So I providing this fix in hope to help some other poor soul, hope it gets included :).

I'm not sure what to do if no applicable schema was found (schema == undefined and return: []). Imo this should also fail, but it could imagine use cases where where you would expect this to pass.

### DIFF
--- a/src/services/jsonSchemaService.ts
+++ b/src/services/jsonSchemaService.ts
@@ -585,14 +585,18 @@ export class JSONSchemaService implements IJSONSchemaService {
 		if (schema) {
 			return this.promise.resolve(jsonDocument.getMatchingSchemas(schema).filter(s => !s.inverted));
 		}
-		return this.getSchemaForResource(document.uri, jsonDocument).then(schema => {
-			if (schema) {
-				return jsonDocument.getMatchingSchemas(schema.schema).filter(s => !s.inverted);
-			}
-			return [];
+		return new this.promise((resolve, reject) => {
+			this.getSchemaForResource(document.uri, jsonDocument).then(schema => {
+				if (schema && schema.errors.length === 0) {
+					return resolve(jsonDocument.getMatchingSchemas(schema.schema).filter(s => !s.inverted));
+				} else if (schema && schema.errors.length > 0) {
+					return reject(new Error('Errors while loading json schema: ' + schema.errors.join()));
+				}
+
+				return resolve([]);
+			});
 		});
 	}
-
 }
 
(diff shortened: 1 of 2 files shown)


KapitanOczywisty:
Seems that other methods, besides `doValidation`, also silently ignores schema errors, so this change would be a bit inconsistent. 
With that, calling `doValidation` (with the same arguments) shows any schema errors with code `ErrorCode.SchemaResolveError` / `768`.

In your case you need complete schema, but if someone is good with partially working schema, for example only with one broken ref for work in progress, full rejection would be an issue. I'd suggest going for `doValidation`.

Trias:
I agree that there are several other parts of the code where something similar could happen. But in those cases, errors are exposed with the `errors` property. Here, no errors are reported.

I would argue that a Promise should not be fulfilled in case of errors, but that's of course a matter of preference and a change would imply major changes in the code base, also incompatible ones.

I guess it's up to @aeschli if he accepts the fix :)

KapitanOczywisty:
> But in those cases, errors are exposed with the `errors` property. Here, no errors are reported.

No, they're not, at least to the outside. Only `doValidation` returns errors, obviously.

Bottom line is that `getMatchingSchemas` is for matching schemas *as is*, not for validation. Partial matches are still useful for colorization etc. What if in future there will be error about some deprecation in parsing schema? Or you're depending on someone's sloppy schema? Killing whole functionality is not a good *fix*.

aeschli:
I imagine `getMatchingSchema` is used for hovers and code completions that work on incomplete content, with warnings and errors to be expected. Throwing an error would make the API useless for these cases.

I'd suggest that we add an optional diagnostics reporter or diagnostics array as argument to getMatchingSchema. We should not introduce a different error type

Trias:
@aeschli @KapitanOczywisty I understand. For those use cases mentioned, returning an error would not make sense.

I think having an error argument would make the API unneccesarily complicated and/or inconsistent and this is not what I am intending. Thanks for your support!

PR #73, opened 2020-09-04
Configuration of severity for schema error

### LINKED ISSUE #22: Configurition of severity for schema error
Hi,
Is there a possibility to add configuration to the severity of schema problems?
Perhaps with the documentSettings.
I would like to define that all schema problems have a severity of error.
Best regards,
Adam

KosrenDQ:
I give it a try :D 

Closes #22

### DIFF
--- a/src/jsonSchema.ts
+++ b/src/jsonSchema.ts
@@ -66,6 +66,7 @@ export interface JSONSchema {
 	suggestSortText?: string;  // VSCode extension
 	allowComments?: boolean; // VSCode extension
 	allowTrailingCommas?: boolean; // VSCode extension
+	severity?: string; // VSCode extension
 }
 
 export interface JSONSchemaMap {
--- a/src/parser/jsonParser.ts
+++ b/src/parser/jsonParser.ts
@@ -379,7 +379,7 @@ function validate(n: ASTNode | undefined, schema: JSONSchema, validationResult: 
 			if (!schema.type.some(matchesType)) {
 				validationResult.problems.push({
 					location: { offset: node.offset, length: node.length },
-					severity: DiagnosticSeverity.Warning,
+					severity: _toDiagnosticSeverity(schema.severity),
 					message: schema.errorMessage || localize('typeArrayMismatchWarning', 'Incorrect type. Expected one of {0}.', (<string[]>schema.type).join(', '))
 				});
 			}
@@ -388,7 +388,7 @@ function validate(n: ASTNode | undefined, schema: JSONSchema, validationResult: 
 			if (!matchesType(schema.type)) {
 				validationResult.problems.push({
 					location: { offset: node.offset, length: node.length },
-					severity: DiagnosticSeverity.Warning,
+					severity: _toDiagnosticSeverity(schema.severity),
 					message: schema.errorMessage || localize('typeMismatchWarning', 'Incorrect type. Expected "{0}".', schema.type)
 				});
 			}
@@ -406,7 +406,7 @@ function validate(n: ASTNode | undefined, schema: JSONSchema, validationResult: 
 			if (!subValidationResult.hasProblems()) {
 				validationResult.problems.push({
 					location: { offset: node.offset, length: node.length },
-					severity: DiagnosticSeverity.Warning,
+					severity: _toDiagnosticSeverity(schema.severity || notSchema.severity),
 					message: localize('notSchemaWarning', "Matches a schema that is not allowed.")
 				});
 			}
@@ -454,7 +454,7 @@ function validate(n: ASTNode | undefined, schema: JSONSchema, validationResult: 
 			if (matches.length > 1 && maxOneMatch) {
 				validationResult.problems.push({
 					location: { offset: node.offset, length: 1 },
-					severity: DiagnosticSeverity.Warning,
+					severity: _toDiagnosticSeverity(schema.severity),
 					message: localize('oneOfWarning', "Matches multiple schemas when only one must validate.")
 				});
 			}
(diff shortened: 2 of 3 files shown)


KapitanOczywisty:
@KosrenDQ  I think that OP wanted setting for all schemas, so I would argue that this closes issue 😉 

As I understand `severity` affect only current schema, but I think additional property which also affects subschemas could be useful. Anyway looks good 👍

aeschli:
I'm not sure how practical it is to have a severity per schema.
- the severity applies to multiple validation checks. E.g. is the type correct, are the number constraints fulfilled. You probably want to express what validation is how severe.
- the schema author needs to annotate each node

Also unclear to me is what it mean to set the severity to ignore or info. Is this the same as removing the constraints? What does an ignore/info mean for `oneOf`

Looking at #22, I believe a simple severity property at the root might be good enough. Or as @KapitanOczywisty suggests, that the severity is valid for all child nodes as well.

The alternative is that each problem gets a 'kind' and there's the possibility to set the severity for each problem kind (for the whole schema)

I think we need some concrete use cases...

KapitanOczywisty:
@aeschli  PR doesn't include `ignore` severity, there is already `ignore` severity in other parts of code to allow trailing comas and comments ;) `errorMessage` also applies to multiple errors, still not to all though. One thing this could be useful is for changing deprecation severity into notice.

KosrenDQ:
Thanks for your Feedback!

@aeschli First, I want to explain my use case. I use a 'schema' to validate a configuration file for my extension. Without the severity settings all problems will be marked as a 'warning'. To avoid an extra validation step of the config file in my extension I want to change the problem severity for my 'schema' from 'warning' to 'error'.
For my use case a simple severity property at the root would be enough. But maybe someone other could need a more granular configuration of the severity :D

@KapitanOczywisty I can add the `ignore` severity :D

KapitanOczywisty:
@KosrenDQ `ignore` part was only the clarification.

So you need only single severity and you'll be adding property to every single schema after this PR? Looks like global `schemaSeverity` is what you need after all.

In this topic I'm thinking to add something like `triggerError`, with that severity could be useful
```js
{
  "if": {
    "anyOf": [
      {
        "required": ["category"],
        "properties": { "category": { "minProperties": 1 } }
      },
      {
        "required": ["product"],
        "properties": { "product": { "minProperties": 1 } }
      }
    ]
  },
  "else": {
    "triggerError": "Provide at least one category or product item",
    "severity": "error"
  }
}
```

BTW in `src/services/jsonValidation.ts` schema resolve errors are fixed to be warning, maybe they should be errors.
https://github.com/microsoft/vscode-json-languageservice/blob/5e62091ee8e84d898a9440b05540dbe9f07e86c8/src/services/jsonValidation.ts#L59-L66

KapitanOczywisty:
In your case would be nice to change diagnostics on the fly, but only other extensions like error lens are seeing changes, in problem tab severity remains unchanged.
```ts
  languages.onDidChangeDiagnostics((event) => {
    for (let uri of event.uris) {
      if (!uri.path.match(/\.json$/)) {
        continue;
      }
      const diagnostics = languages.getDiagnostics(uri);
      for (let single of diagnostics) {
        if (single.severity !== DiagnosticSeverity.Warning || single.source === "myExtension") {
          continue;
        }
        single.severity = DiagnosticSeverity.Error;
        single.source = "myExtension";
      }
    }
  });
```

KosrenDQ:
Thank you for the hint, to change the diagnostics in my extension will resolve my problems. 

But why should I add the property to all `schemas` after my PR? 
```
function _toDiagnosticSeverity(level: string | undefined): DiagnosticSeverity {
    switch (level) {
	case 'error': return DiagnosticSeverity.Error;
	case 'warning': return DiagnosticSeverity.Warning;
	case 'information': return DiagnosticSeverity.Information;
	case 'hint': return DiagnosticSeverity.Hint;
	default: return DiagnosticSeverity.Warning;
    }
}
```
Without the property the diagnostic severity will be set the default `warning`. Do I miss something?

KapitanOczywisty:
> First, I want to explain my use case. I use a `schema` to validate a configuration file for my extension. Without the severity settings all problems will be marked as a `warning`. To avoid an extra validation step of the config file in my extension I want to change the problem severity for my `schema` from `warning` to `error`.
> For my use case a simple severity property at the root would be enough. But maybe someone other could need a more granular configuration of the severity :D

As I understand you want to change every single schema / subschema to emit errors instead of warnings, but `severity` property changes only one schema/node, but maybe I have wrong picture of your use case.

aeschli:
@KosrenDQ With your original PR. if a schema author wanted all problems to be errors, he would have to put `"severity": "error"` at on every sub scheme.
```{
    "type": "object",
    "required": [
        "type"
    ],
    "properties": {
        "type": {
            "type": "string",
            "const": "Type_B",
            "severity": "error"
        },
        "age": {
            "type": "number",
            "maximum": 4,
            "severity": "error"
        }
    },
    "additionalProperties": false,
    "severity": "error"
}
```

aeschli:
in #22, @adamYzamri suggests to add a severity parameter to the `DocumentSettings` of the `doValidation` call. So the user of the JSON library decides what the severity of the created Diagnostics are, and in the case of VSCode it will be a user setting that defines the severity of JSON schema validation problems.

We currently use `warning` as default as we thought that validations were less severe that syntax errors. But we probably could make 'error' the default.

KapitanOczywisty:
> in #22, @adamYzamri suggests to add a severity parameter to the `DocumentSettings` of the `doValidation` call. So the user of the JSON library decides what the severity of the created Diagnostics are, and in the case of VSCode it will be a user setting that defines the severity of JSON schema validation problems.

In that case you can loop through returned diagnostics and change severity without any problems. This would be rather small qol update rather than new feature. With `DocumentSettings` severity per type would make sense, but all of this doesn't improve anything for vscode user and for extension you have to be middle man and disable somehow default validation.

KosrenDQ:
Now I see the point with `schemas` and `subSchemas`, sorry I misunderstood you. But yes, I would add the `severity` to each `schema`.
However, I think to use the `DocumentSettings` is a much more elegant way to achieve a configurable `severity`. On the other hand to you change the default `severity` to `error` would be fine for me, too. :D

aeschli:
@KosrenDQ Ok, let's do that first. Do you want to look at a PR or should I do it?

KosrenDQ:
@aeschli I can do a new PR, should I change the `default severity` or add a `DocumentSetting`?

aeschli:
Let's add a `schemaValidation: SeverityLevel` to `DocumentLanguageSettings`.
If set to ignore, schema validation can be skipped.

We can also change the default, or we keep it compatible and I set severity to error in the JSON language server.

KosrenDQ:
Ok then i will close this PR, thank you!

PR #74, opened 2020-09-09
Add severity DocumentSetting

### LINKED ISSUE #22: Configurition of severity for schema error

KosrenDQ:
As discussed in #73, add a `schame severity` as a `DocumentSetting`.  
Should close #22.

### DIFF
--- a/src/jsonLanguageTypes.ts
+++ b/src/jsonLanguageTypes.ts
@@ -134,6 +134,11 @@ export interface DocumentLanguageSettings {
      * The default severity level for schema validation is 'warning'. 
 	 */
 	schemaValidation?: SeverityLevel;
+
+	/**
+	 * The severity of schema resolving. If not set, schema resolving errors will be reported as 'warning'.
+	 */
+	schemaRequest?: SeverityLevel;
 }
 
 export interface SchemaConfiguration {
--- a/src/jsonLanguageTypes.ts
+++ b/src/jsonLanguageTypes.ts
@@ -128,6 +128,11 @@ export interface DocumentLanguageSettings {
 	 * The severity of reported trailing commas. If not set, trailing commas will be reported as errors.
 	 */
 	trailingCommas?: SeverityLevel;
+
+	/**
+	 * The severity of schema validation. If se to 'ignore', schema validation will be skipped. 
+	 */
+	schemaValidation?: SeverityLevel;
 }
 
 export interface SchemaConfiguration {
--- a/src/jsonLanguageTypes.ts
+++ b/src/jsonLanguageTypes.ts
@@ -130,7 +130,8 @@ export interface DocumentLanguageSettings {
 	trailingCommas?: SeverityLevel;
 
 	/**
-	 * The severity of schema validation. If se to 'ignore', schema validation will be skipped. 
+	 * The severity of schema validation. If set to 'ignore', schema validation will be skipped. If this property is not set, the default severity level will be used.
+     * The default severity level for schema validation is 'warning'. 
 	 */
 	schemaValidation?: SeverityLevel;
 }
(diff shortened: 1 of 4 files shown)


src/jsonLanguageTypes.ts
```diff
@@ -128,6 +128,11 @@ export interface DocumentLanguageSettings {
 	 * The severity of reported trailing commas. If not set, trailing commas will be reported as errors.
 	 */
 	trailingCommas?: SeverityLevel;
+
+	/**
+	 * The severity of schema validation. If se to 'ignore', schema validation will be skipped.
```
aeschli, line 133: We should also define what the default is. I suggest, for backward compatibility, 'warning'.

Also there's a `t` missing in `set`

aeschli (changes requested):

KapitanOczywisty:
@aeschli @KosrenDQ 
1. TBH name is not the best and there is already `validate` in `LanguageSettings`
https://github.com/microsoft/vscode-json-languageservice/blob/d1278297336d6f220140cf9636c234f80070b480/src/jsonLanguageTypes.ts#L102-L106
It would make sense to use `validate?: boolean` in `DocumentLanguageSettings` and add  `errorSeverity` (or so), maybe even add it to both of them.

2. Schema errors are not affected by change, I'm not sure if they should be.
https://github.com/microsoft/vscode-json-languageservice/blob/d1278297336d6f220140cf9636c234f80070b480/src/services/jsonValidation.ts#L60-L67

3. I think error severity should be set in code rather than replaced in loop. For example `deprecationMessage` probably should always be `Information` or `Warning`. (btw. it's missing `DiagnosticTag.Deprecated`)
https://github.com/microsoft/vscode-json-languageservice/blob/d1278297336d6f220140cf9636c234f80070b480/src/parser/jsonParser.ts#L547-L553

The need of any other severity for any message in the future, will make this change obsolete. I don't have much time lately, but soon I'll make PR with `triggerError`. Example/use case:
```json
{
  "if": {
    "anyOf": [
      { "required": ["category"], "properties": { "category": { "minProperties": 1 } } },
      { "required": ["product"], "properties": { "product": { "minProperties": 1 } } }
    ]
  },
  "else": {
    "triggerError": "Provide at least one category or product item",
    // or
    "triggerError": {
      "severity": "warning",
      "message": "Provide at least one category or product item"
    }
  }
}
```

4. I think that errors caused by schema should also have error code. Currently only enum uses `ErrorCode.EnumValueMismatch`, which is a bit weird.

Sorry for a bit of "side topics", but errors needs some changes and this seems to be a good place to bring this in.

aeschli:
`validate` is to turn off both syntax and semantic errors and the new setting is just for the semantic validation, so I think they play well together.
I agree that deprecation should remain a warning. To distinguish these we need to mark the problem with a new error code. I agree that we should have error codes for all kind of errors.

Maybe we just remove `IProblem.severity` for now,

Yes, schema load errors should have a separate severity.
`schemaRequests?: SeverityLevel`

aeschli:
@KapitanOczywisty Before you start on a PR for `triggerError`, lets discuss it, I need to better understand the feature.

KapitanOczywisty:
@aeschli this is probably 5 lines of code and wouldn't be any bad (in terms of wasted time) if this was rejected, but if you want we can discuss.

ATM if you want custom validation rules you have to make extension, but some of them could be written in schema using `if`/`then`/`else`, like example above. Currently you could use `deprecationMessage` to achieve this, however it has too big diagnostic range and `DiagnosticTag` might be added someday (as it probably should). Also `$data` is supported by some packages, afik it's only proposal, but still could make new opportunities for extended validation. Not sure what I could add more.

src/jsonLanguageTypes.ts
KosrenDQ: I have improved the description :D

KosrenDQ:
I have added the attribute `schemaRequest` for changing the severity of the schema loading errors.

KosrenDQ:
Any chance this will be merged or should I change something? :D

aeschli:
Thanks @KosrenDQ !

aeschli:
I'll make a change so that deprecation stays a warning

PR #77, opened 2020-10-01
Convert "hover" action to return MarkedContent instead of MarkedString

### LINKED ISSUE #53: Add tests for hover contributions
The current tests use schema contributions to test various schema features.

However there are no tests for hover `contributions`, which poses a problem if one wanted to write alternative hover providers.

The relevant code is here:
https://github.com/microsoft/vscode-json-languageservice/blob/fabbc8069d2b03b0ac73c1f710be4a5535df41d5/src/services/jsonHover.ts#L54-L60

apexskier:
`MarkedString` is deprecated by the LSP, and an editor I'm working in isn't escaping it properly. This updates the hover action to return `MarkedContent` instead of `MarkedString[]`. As part of this, it can now return plain text using `MarkedKind.PlainText`, which avoids some markdown escaping. I've maintained this package's custom extension to Schemas to support markdown content documentation though.

Additionally, I've added some basic tests for hover contributions (resolves #53) and replaced the deprecated `assert.deepEqual` method in these tests.

### DIFF
--- a/src/jsonContributions.ts
+++ b/src/jsonContributions.ts
@@ -2,10 +2,10 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
-import { Thenable, MarkedString, CompletionItem } from './jsonLanguageService';
+import { Thenable, CompletionItem, Hover } from './jsonLanguageService';
 
 export interface JSONWorkerContribution {
-	getInfoContribution(uri: string, location: JSONPath): Thenable<MarkedString[]>;
+	getInfoContribution(uri: string, location: JSONPath): Thenable<Hover["contents"]>;
 	collectPropertyCompletions(uri: string, location: JSONPath, currentWord: string, addValue: boolean, isLast: boolean, result: CompletionsCollector): Thenable<any>;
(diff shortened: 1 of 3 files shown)


src/jsonContributions.ts
```diff
@@ -2,10 +2,10 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
-import { Thenable, MarkedString, CompletionItem } from './jsonLanguageService';
+import { Thenable, CompletionItem, Hover } from './jsonLanguageService';
 
 export interface JSONWorkerContribution {
-	getInfoContribution(uri: string, location: JSONPath): Thenable<MarkedString[]>;
+	getInfoContribution(uri: string, location: JSONPath): Thenable<Hover["contents"]>;
```
apexskier, line 8: This is a wider type, so is fully backwards compatible.

aeschli:
Thanks a lot!

We have this client capability that is passed when service is created: https://github.com/microsoft/vscode-json-languageservice/blob/9ac8f6ebfa797bfb5a0a1398a3b8b605b1c7ce1f/src/jsonLanguageTypes.ts#L237

It lets the API user control whether the hover can use Markdown. Now with the MarkedContent we should probably observe that capability.

PR #78, opened 2020-10-24
Improve README

bogem:
* Format with Prettier
* Get rid of repeating point
* Add missing dots
* Improve some points
* Improve "Development" section


aeschli:
Thanks a lot @bogem !

Issue #80, opened 2020-11-20
use real glob instead of simpleRegex


Trias:
I'm not sure if this is documented, but I was under the impression that the syntax used when providing [fileMatch](https://github.com/microsoft/vscode-json-languageservice/blob/36f8016a33df9eff3b6edd584605aea349b73745/src/jsonLanguageTypes.ts#L154)-Patterns for schemas ist a "glob"-Pattern. But it appers it is a "Simpleregex" defined by replacing certain characters as defined here: https://github.com/microsoft/vscode-json-languageservice/blob/master/src/utils/strings.ts#L34-L36

The key difference i noticed is that `/**/` in a glob pattern matches **zero** or more folders, whereas in "simpleRegex" it matches **one** or more nested folders.

There is a package called [glob-to-regex](https://www.npmjs.com/package/glob-to-regexp) which appears to do the conversion respecting "globstar".

I'm happy to do the pull-request if you would approve this change.

aeschli:
As written in the API comment, currently we only support '*', not glob
I'm not opposed to allow glob as well but ideally implement it ourselves inside json-languageservice so I don't need to spend any time with legal approvals.

Trias:
Understandable. Maybe we can use this code: https://github.com/microsoft/vscode/blob/master/src/vs/base/common/glob.ts?

Not sure tough how to integrate this code into into this library.

I realized that allowing glob pattern (obviously) is a breaking change, so probably we need to provide an alternative option in addition to fileMatch.

mifieldxu:
hello! i'm new here. it turned out that the vscode-common repo doesn't actually have glob.ts in it. i implemented a solution to this with the [vscode-anymatch](https://www.npmjs.com/package/vscode-anymatch) package, which is just a fork of [anymatch](https://github.com/micromatch/anymatch), but since it's a dependency for [vscode-chokidar](https://github.com/paulmillr/chokidar), i assume it's ok to use. 

i also wrote some additional tests to make sure it works as expected, presuming that fileMatch is intended to be glob patterns. 

i went ahead and submitted a pull request, but do feel free to do whatever you'd like!

ssbarnea:
Do we have any chance to get a some progress on this? Currently due to the very limited implementation it is impossible to correctly select schemas for some real world use-cases.

I am wondering it would be ok to just assume that `**/ -> .*`  and `* -> [^\/]*`. Two stars should cross directory boundaries but one star should only match anything but the directory separator.

[...]

PR #83, opened 2020-12-14
implement a solution for issue #80, using vscode-anymatch

mifieldxu:
please see comments on the [issue page](https://github.com/microsoft/vscode-json-languageservice/issues/80)


src/services/jsonSchemaService.ts
```diff
@@ -69,41 +74,44 @@ export interface ISchemaHandle {
...
 
 	constructor(pattern: string[], uris: string[]) {
-		this.patternRegExps = [];
-		this.isInclude = [];
+		this.anymatchTesters = [];
 		try {
-			for (let p of pattern) {
-				const include = p[0] !== '!';
+			for (let patternString of pattern) {
+				const include = pattern[0] !== BANG;
```
Trias, line 89: this does seem like a bug... was the intention here to check `patternString[0]` ?

aeschli:
`vscode-anymatch` has been deprecated in the emantime. I suggest to use `minimatch` instead.

PR #88, opened 2021-02-15
add deprecation annotation from json schema draft 2019-09

### LINKED ISSUE #87: Deprecation in Validations
Looking at the code it does not seem to be that case that "deprecationMessage" or "deprecated: true" (new property in JSON Schema 2019-09) are supported in Diagnostics.

LSP supports this since version 3.15 as a "DiagnosticTag", see https://microsoft.github.io/language-server-protocol/specification#diagnostic

I would be interested in providing a pull request, just want to make sure in advance if you are interested and/or have any pointers.

Trias:
fixes #87 

before:
2021-02-15 12_17_38-● settings json - vscode-json-languageservice - Visual Studio Code

after:
2021-02-15 12_04_16-settings json - Code - OSS Dev

I've added the deprecation tag to the language service by implementing the "depreated" property from json schema draft 2019-09. The custom "deprecationMessage" also works. Deprecated symbols are now rendered with a strikethrough instead of a warning.

**open questions:** 
I've added a new language snippet "deprecationMessage", I don't know how the translation process is going to work.

### DIFF
--- a/src/jsonLanguageTypes.ts
+++ b/src/jsonLanguageTypes.ts
@@ -7,10 +7,10 @@ import { JSONWorkerContribution, JSONPath, Segment, CompletionsCollector } from 
 import { JSONSchema } from './jsonSchema';
 import {
 	Range, TextEdit, Color, ColorInformation, ColorPresentation, FoldingRange, FoldingRangeKind, MarkupKind, SelectionRange,
-	Diagnostic, DiagnosticSeverity,
+	Diagnostic, DiagnosticSeverity,	DiagnosticTag,
 	CompletionItem, CompletionItemKind, CompletionList, Position,
 	InsertTextFormat, MarkupContent,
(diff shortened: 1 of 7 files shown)


Trias:
I've refined support for deprecations a bit. Before, deprecations would invalidate a schema, which they are not supposed to.

I hope i didn't miss anything.

the example above is now rendered slightly differently: 
2021-02-17 12_32_50-settings json - Code - OSS Dev

aeschli:
Thanks a lot for the PR. I think its good, but I don't like that the deprecationResult is now everywhere.

The reason for this was the 'hasProblems` check, right? What about just changing that one to hasErrors and have it ignore problems with severity warning and less?

Trias:
yeah that would make more sense.... i'm not sure why i duplicated it. i will look into it :)

Trias:
i'm no longer working on this. happy if someone picks it up though :)

PR #93, opened 2021-03-11
improved glob matching using minimatch

mifieldxu:
This is essentially the pull request as [#83](https://github.com/microsoft/vscode-json-languageservice/pull/83), which was closed due to branch renaming.

The solution now uses minimatch, at the suggestion of @aeschli. All tests pass as far as I can tell.


aeschli:
Thansk @mifieldxu and @ssbarnea !

PR #95, opened 2021-03-16
Upgrade dependencies

ssbarnea:
Raise minimal npm version to 7 in order to assure that we can use
the newer lockfileVersion version and we do not endup switching back
and forth between versions.

This includes dependency updates as done by `npm install`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,31 +1,2435 @@
 {
   "name": "vscode-json-languageservice",
   "version": "4.0.2",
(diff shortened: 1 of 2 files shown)


ssbarnea:
@aeschli Please review. This one is preparatory towards the one that is adding globbing. the lock version changed and any operation with npm upgrades the lock file, better to do that outside other changes.

aeschli:
Thanks @ssbarnea !

PR #97, opened 2021-03-30
Allow case-insensitive modifier at the very beginning of the pattern

### LINKED ISSUE #96: Add support for passing regex flags
The following works in [this](https://www.jsonschemavalidator.net/) online JSON schema validator:

```json
{
  "type": "string",
  "pattern": "(?i)^alt\\+C$"
}
```

But from Monaco I'm getting this error: `SyntaxError: Invalid regular expression: /(?i)^alt\+C$/: Invalid group.`

It looks like regex flags can't be passed at all with this library:

https://github.com/microsoft/vscode-json-languageservice/blob/07fec3faafa3f32915dcd6fcc6999a91ad113f88/src/parser/jsonParser.ts#L647-L648

IMHO this is something pretty necessary, one can't just compile away the "i" flag for example, the combinatorics would explode in some cases.

KapitanOczywisty:
Feature + tests. Ready for review @aeschli

Fixes #96

Edit: Changed to single quotes for consistency
Edit2: Also check for fail with modifier

### DIFF
--- a/src/parser/jsonParser.ts
+++ b/src/parser/jsonParser.ts
@@ -6,6 +6,7 @@
 import * as Json from 'jsonc-parser';
 import { JSONSchema, JSONSchemaRef } from '../jsonSchema';
 import { isNumber, equals, isBoolean, isString, isDefined } from '../utils/objects';
+import { extendedRegExp } from '../utils/strings';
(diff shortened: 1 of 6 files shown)


aeschli:
Thanks @KapitanOczywisty !

Issue #103, opened 2021-08-24
[json] support unicode string pattern "^\\p{L}$"


sourabh112:
Referring microsoft/vscode#131207

sourabh112:
@aeschli As I said in [this comment](https://github.com/microsoft/vscode/issues/131207#issuecomment-903798575) this problem is occurring because of the usage of javascript RegExp. We can solve it by Using xRegExp node module, As it supports Unicode.

aeschli:
I would like a solution that does not add any dependencies and doesn't add much to the size of vscode-json-languageservice. Tha way we stay clear of licensing and security issues and are in control of the code that get shipped along with `vscode-json-languageservice`.
`xregexp` is ~ 130kb of source, almost the size of the vscode-json-languageservice.

sourabh112:
So, I found this [script](https://difnet.com.br/opensource/unicode_hack.js) which is basically converting Unicode to normal/conventional regular expressions. I have tested it on strings of the original issue and it seems to be working great.
![image](https://user-images.githubusercontent.com/66176305/130616372-bb341bdf-fb90-4849-ab34-2d635c299f40.png)

sourabh112:
@aeschli Above script is the best way. I found for converting unicodes to regexp acceptable regular expressions. I think we can use it as it's size is low around 23-24 kb or I can implement this function on my own is a new js file which we can add in the repo (but it might not be as optimized as this.).

sourabh112:
Hi @aeschli , The string pattern that I'm receiving in extendedregExp is "^\p{L}$" not "^\\p{L}$". Can you please explain it why this happening?

sourabh112:
@aeschli I have submitted the PR. Can you please review it.

PR #104, opened 2021-08-29
Adding Unicode support

sourabh112:
Fixes Issue: #103 microsoft/vscode#131207
Adding a file `Unicode_support.ts` in order to add support for Regular Expressions which uses Unicode, in JSON Schema. The same file can be used anywhere in place of the normal regexp function of javascript to add Unicode Support.


sourabh112:
@aeschli The Unicode file I have added is quite a bit different from this [script](https://difnet.com.br/opensource/unicode_hack.js), I talked about it in my [earlier comment](https://github.com/microsoft/vscode-json-languageservice/issues/103#issuecomment-905512203). But the basic idea is the same from that code. As I don't know much about copyright, etc. so, I have kept that first 3 lines intact. I will remove them if they are not necessary.

jeanp413:
@sourabh112 @aeschli is there a reason that is blocking `vscode-json-languageservice` to update it's tsconfig file and use `"ES2018.Regexp"` as vscode https://github.com/microsoft/vscode/blob/main/src/tsconfig.base.json.
Doing that the fix would be trivial and just need to add the 'u' option when creating the regex.

sourabh112:
@jeanp413 What you are suggesting is already present in `ES2019.Regexp` and for that, to work we have to ask the user to add /u after every Unicode which is technically not wrong but is still a workaround because if you read the original issue `pattern: "^\\p{L}$"` should pass `example: "a"` but it doesn't because they didn't added /u at the back. Now If we add /u at the back the pattern will become `pattern: "^\/\\p{L}\/u$"` it will pass `example: "a"` but the online JSON validator fails it. ( I'm not very good at explaining. I hope you get my point).

The second way to solve it using regexp would be by using a TS file to update the given pattern and add / before Unicode and /u after the Unicode. and It would be the same as my implementation.

sourabh112:
I might be wrong. So @jeanp413 and @aeschli please confirm.

jeanp413:
I was thinking that unicode support should always be enabled by default. 
```ts
 	if (startsWith(pattern, '(?i)')) { 
 		return new RegExp(pattern.substring(4), 'iu'); 
 	} else { 
 		return new RegExp(pattern, 'u'); 
 	} 
```

sourabh112:
@jeanp413 I don't know why I didn't think of that before. yes, you are right this will do it.

KapitanOczywisty:
Both `i` and `u` can lead to performance decrease in some cases, I think it's not a good idea to add any of them to all patterns. It's probably better idea to add full support for the first `(?...)` at the very beginning as it was started in https://github.com/microsoft/vscode-json-languageservice/issues/96#issuecomment-809683294 / #97

sourabh112:
@KapitanOczywisty We can add my file for the `(?...)` case and leave `u` for the rest.

sourabh112:
I don't know whether it will increase the performance or not.

sourabh112:
@KapitanOczywisty @aeschli Can you please review the changes?

aeschli:
@sourabh112 Can you add some tests?

sourabh112:
Sure, I will add some. Directed towards my changes.

KapitanOczywisty:
Since JS (starting from ES6) is natively supporting `\p{..}` sequences using `u` modifier using third party script is at least questionable.

I've checked few libraries and inline modifier `(?u)` is not supported, so this cannot be solved as with case-insensitive.

Enabling `u` flag globally causes some [side effect](https://mathiasbynens.be/notes/es6-unicode-regex). However seems like other libraries are doing exactly that. And on top of that specs starting from [`2020-12`](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-6.4) requires `u` flag.

>  Regular expressions SHOULD be built with the "u" flag (or equivalent)
>  to provide Unicode support, or processed in such a way which provides
>  Unicode support as defined by ECMA-262.

https://ajv.js.org/options.html#unicoderegexp - ajv has option to use `u` flag - enabled by default
https://github.com/tdegrunt/jsonschema - allways `u`
both doesn't support case insensitive modifier
https://www.jsonschemavalidator.net/ - supports unicode and inline modifiers (always enabled in .NET)

@sourabh112 `u` flag is probably desired solution after all.

@aeschli Could there be any problems for vscode itself with `u` flag?

sourabh112:
@KapitanOczywisty I will add some test cases for now. And Will wait for @aeschli Confirmation for what I should do `u` or my script.

aeschli:
@KapitanOczywisty Thanks for the pointers. https://mathiasbynens.be/notes/es6-unicode-regex is very interesting.

I think we should follow the recommendation and switch to /u. There are some changes as described, but for our usage here (generate validation errors in VS Code) we can accept this and don't need to worry about backward compatibility.

sourabh112:
@aeschli ok, I will remove the script and add `u` instead. And I'm facing some problems in adding test cases. So, I might take some time but I can do it.

PR #105, opened 2021-09-05
fix: clear `JsonSchemaService` local single resource cache on schema change

### LINKED ISSUE #106: Dynamic schema does not update when changing for current document and not using $schema property
I am using a dynamic schema in an vscode extension that implements an `vscode.TextDocumentContentProvider` that fires an `onDidChange` event on schema change. This in turn sends a schema content change notification (`json/schemaContent`) to the language server, which calls this language service's `languageService.resetSchema(uri)`.

That's all good, but unfortunately the `JsonSchemaService` does not clear its current document cache in this scenario. So this means dynamic schema changes to the current viewing file do not cause any schema changes (In my scenario, the user specifies plugins in the JSON file and based on what plugins they specify the configuration changes).

I've opened #105 to resolve this issue.

Workarounds for user:

1. Change the JSON plugin settings so it clears all the cached schemas.
2. View another JSON file with a schema then switch back to the other file.

dsherret:
This local cache was not being cleared on a resource change.

Closes #106

### DIFF
--- a/src/services/jsonSchemaService.ts
+++ b/src/services/jsonSchemaService.ts
@@ -273,6 +273,9 @@ export class JSONSchemaService implements IJSONSchemaService {
 	}
 
 	public onResourceChange(uri: string): boolean {
+		// always clear this local cache when a resource changes
+		this.cachedSchemaForResource = undefined;
+
 		let hasChanges = false;
 		uri = normalizeId(uri);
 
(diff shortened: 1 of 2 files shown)


src/services/jsonSchemaService.ts
```diff
@@ -273,6 +273,9 @@ export class JSONSchemaService implements IJSONSchemaService {
 	}
 
 	public onResourceChange(uri: string): boolean {
+		// always clear this local cache when a resource changes
+		this.cachedSchemaForResource = undefined;
```
dsherret, line 277: Perhaps this could be more fine targeted, but I'm not sure it's worth the complexity.

src/test/schema.test.ts
```diff
@@ -1176,6 +1176,58 @@ suite('JSON Schema', () => {
...
+				a: {
+					type: 'number'
+				}
+			}
+		};
+		ls.resetSchema(schemaUri);
+
+		// now ensure validation occurs with the new schema
+		validation = await ls.doValidation(testDoc.textDoc, testDoc.jsonDoc);
+		assert.deepStrictEqual(validation.map(v => v.message), []);
```
dsherret, line 1227: Previously this would use the resource on `this.cachedSchemaForResource` instead of the latest cache.

aeschli:
Thanks a lot @dsherret !

Issue #102, opened 2021-07-19
Consider using the official test suite?


Relequestual:
Hello!

Given the importance of this module when using JSON Schema, I wonder, why not use the official test suite?

I did read you don't want external dependencies. Currently we do not offer a maintained version of the test suite as an npm release. We usually suggest people use git submodules to capture the tests they need.

I do see you have tests beyond validation. The official test suite currently only covers validation.

If it's difficult to justify time to do this, I may be able to help you find some additional justification.
Making the change is beyond the time I can offer currently.

Trias:
(not the maintainer) As far as i know the module isn't in full compliance with the test suite yet. [This issue](https://github.com/microsoft/vscode-json-languageservice/issues/55) seems to be the most prominent which would fail many tests of the official test suite.

Relequestual:
Thanks for linking @Trias !

The [Backlog](https://github.com/microsoft/vscode-json-languageservice/milestone/4) is quite small, so we can hope!

Relequestual:
> Can you point me to the official test suite?

You can find it here: https://github.com/json-schema-org/JSON-Schema-Test-Suite
Apologies for the long delay!

PR #107, opened 2021-09-29
Fix $ref to $id

### LINKED ISSUE #55: $ref should accept $id references
As part of JSONSchema Draft 6, $ref can point to a definition defined with an `$id` (or just `id`, spec seems vague if its `$id` or `id` in v6) -- it would be wonderful if we could get support for that here.

For instance, in definitions we will be able to use `{"references": { "$id": "#test", ... }}` and in our core properties, refer to it via `{"properties": {"$ref": "#test", ...}}`

Thanks!

mati-o:
Hi,
I am attempting to fix #55 and microsoft/vscode#92042. I have implemented a simple workaround to detect whether a `$ref` points to an `$id` and then I attempt to find the definition behind that `$id`.

I have implemented it inside `findSection` of `resolveSchemaContent` as I though it would be the simplest low-level solution. I have also added some unit tests. Everything works just as before - all tests pass!

I'll be more than happy to see this PR accepted, let me know if anything else is required.

### DIFF
--- a/src/services/jsonSchemaService.ts
+++ b/src/services/jsonSchemaService.ts
@@ -123,21 +123,23 @@ class FilePatternAssociation {
 	}
 }
 
-type SchemaDependencies = { [uri: string]: true };
+type SchemaDependencies = Set<string>;
 
 class SchemaHandle implements ISchemaHandle {
 
-	public url: string;
-	public dependencies: SchemaDependencies;
+	public readonly url: string;
+	public readonly dependencies: SchemaDependencies;
+	public readonly anchors: Map<string, JSONSchema>;
 
 	private resolvedSchema: Thenable<ResolvedSchema> | undefined;
 	private unresolvedSchema: Thenable<UnresolvedSchema> | undefined;
-	private service: JSONSchemaService;
+	private readonly service: JSONSchemaService;
 
 	constructor(service: JSONSchemaService, url: string, unresolvedSchemaContent?: JSONSchema) {
 		this.service = service;
 		this.url = url;
-		this.dependencies = {};
+		this.dependencies = new Set();
+		this.anchors = new Map();
 		if (unresolvedSchemaContent) {
 			this.unresolvedSchema = this.service.promise.resolve(new UnresolvedSchema(unresolvedSchemaContent));
 		}
@@ -153,7 +155,7 @@ class SchemaHandle implements ISchemaHandle {
 	public getResolvedSchema(): Thenable<ResolvedSchema> {
 		if (!this.resolvedSchema) {
 			this.resolvedSchema = this.getUnresolvedSchema().then(unresolved => {
-				return this.service.resolveSchemaContent(unresolved, this.url, this.dependencies);
+				return this.service.resolveSchemaContent(unresolved, this);
 			});
 		}
 		return this.resolvedSchema;
@@ -162,7 +164,8 @@ class SchemaHandle implements ISchemaHandle {
 	public clearSchema() {
 		this.resolvedSchema = undefined;
 		this.unresolvedSchema = undefined;
-		this.dependencies = {};
+		this.dependencies.clear();
+		this.anchors.clear();
 	}
 }
 
@@ -286,7 +289,7 @@ export class JSONSchemaService implements IJSONSchemaService {
 			const curr = toWalk.pop()!;
 			for (let i = 0; i < all.length; i++) {
 				const handle = all[i];
-				if (handle && (handle.url === curr || handle.dependencies[curr])) {
+				if (handle && (handle.url === curr || handle.dependencies.has(curr))) {
 					if (handle.url !== curr) {
 						toWalk.push(handle.url);
 					}
@@ -401,7 +404,7 @@ export class JSONSchemaService implements IJSONSchemaService {
 		);
 	}
 
-	public resolveSchemaContent(schemaToResolve: UnresolvedSchema, schemaURL: string, dependencies: SchemaDependencies): Thenable<ResolvedSchema> {
+	public resolveSchemaContent(schemaToResolve: UnresolvedSchema, handle: SchemaHandle): Thenable<ResolvedSchema> {
 
 		const resolveErrors: string[] = schemaToResolve.errors.slice(0);
 		const schema = schemaToResolve.schema;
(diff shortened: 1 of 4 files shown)


KapitanOczywisty:
`$id` could be used in any place, not only in the definitions, even after referencing node. Allowing to use `$id` only in definitions is a step in the right direction, but still very limited upgrade.
```json
{
  "type": "object",
  "properties": {
    "p1": {
      "$ref": "#hello"
    },
    "p2": {
      "$ref": "#world"
    },
    "p3": {
      "id": "#hello",
      "type": "string",
      "const": "hello"
    },
    "p4": {
      "type": "object",
      "properties": {
        "deep": {
          "$id": "#world",
          "type": "string",
          "const": "world"
        }
      },
      "additionalProperties": false
    }
  },
  "additionalProperties": false
}
```

Also there should be a check with node referencing their parent, to test recursion:
```json
{
  "type": "object",
  "definitions": {
    "foo": {
      "id": "#foo",
      "type": "object",
      "properties": {
        "bar": {
          "type": "string",
          "const": "hello"
        },
        "foo": {
          "$ref": "#foo"
        }
      },
      "additionalProperties": false
    }
  },
  "properties": {
    "foo": {
      "$ref": "#foo"
    }
  },
  "additionalProperties": false
}
```
```json
{
  "foo": {
    "foo": {
      "foo": {
        "bar": "hello"
      }
    }
  }
}
```

And lastly I'd add some test when `$ref` is referencing not existing `$id` to test correct error handling.

mati-o:
I'll take a look later, it won't hurt to add that test. As far as supporting `$id` everywhere, while true, I have supported `definitions` since afaik it's a common practice. I'll try to see how it is possible to implement with minimal performance penalty, probably like you said - collect them during resolution. Hope it won't be complicated 😅

src/services/jsonSchemaService.ts
```diff
@@ -433,20 +433,54 @@ export class JSONSchemaService implements IJSONSchemaService {
...
-				for (const key in section) {
-					if (section.hasOwnProperty(key) && !target.hasOwnProperty(key)) {
-						(<any>target)[key] = section[key];
-					}
-				}
+				merge(target, section);
 			} else {
 				resolveErrors.push(localize('json.schema.invalidref', '$ref \'{0}\' in \'{1}\' can not be resolved.', path, sourceURI));
 			}
 		};
```
mati-o, line 452: I have refactored `merge` to `mergeByJsonPointer` which calls `findSection` and the actual `merge` which just copies the properties over. I did it since when I resolve a subschema, I don't need to look for it since I have the object in hand, but merge is still required.

src/services/jsonSchemaService.ts
```diff
@@ -433,20 +433,54 @@ export class JSONSchemaService implements IJSONSchemaService {
...
+				if (!pendingSubSchemas[uri]) {
+					pendingSubSchemas[uri] = [];
+				}
+
+				// Remember the target to merge later once resolved
+				pendingSubSchemas[uri].push(target);
+				return false; // return failure - merge didn't occur
+			}
+		};
+
```
mati-o, line 483: This is the heart of the solution. It basically keeps two maps for subschema references.  One for those subschemas that we may or may not encounter later and the latter for those subschemas we have resolved. 

The `tryMergeSubSchema` merges already resolved subschemas into the targets that `$ref` them, or keeps a record of the target object against the full URI of the subschema for later use.

src/services/jsonSchemaService.ts
```diff
@@ -527,17 +585,52 @@ export class JSONSchemaService implements IJSONSchemaService {
...
+						while (pendingSubSchemas[fullId].length) {
+							const target = pendingSubSchemas[fullId].shift();
+							merge(target!, next);
+						}
+
+						delete pendingSubSchemas[fullId];
+					}
+				}
+			};
+
```
mati-o, line 617: This function resolves the actual subschema, whenever it encounters an `$id` or `id` in an object. It keeps a record of the full URI and the object and resolves any pending requests for that subschema.

src/services/jsonSchemaService.ts
```diff
@@ -459,8 +493,23 @@ export class JSONSchemaService implements IJSONSchemaService {
...
+					// This is a reference to a subschema
+					const fullId = reconstructRefURI(uri, refSegment);
+
+					if (!tryMergeSubSchema(fullId, node)) {
+						// We weren't able to merge currently so we'll try to resolve this schema first to obtain subschemas
+						// that could be missed
+						externalLinkPromise = resolveRefs(unresolvedSchema.schema, unresolvedSchema.schema, uri, referencedHandle.dependencies);
+					}
+				}
+				return externalLinkPromise.then(() => resolveRefs(node, unresolvedSchema.schema, uri, referencedHandle.dependencies));
```
mati-o, line 512: In case we have referenced a subschema in an external schema i.e `schema1` -> `$ref` -> `schema2#foo`, we'll need to find `schema2`'s `#foo` so we trigger `schema2` resolution eagerly using that `externalLinkPromise`.

mati-o:
I think I have figured this out! See the comments in the code too, hope it is clear enough. Let me know if more tests are needed

src/test/schema.test.ts
```diff
@@ -273,6 +273,243 @@ suite('JSON Schema', () => {
...
+		});
+
+	});
+
+	test('Resolving $refs to external $ids', async function () {
+		const service = new SchemaService.JSONSchemaService(newMockRequestService(), workspaceContext);
+		service.setSchemaContributions({
+			schemas: {
+				"https://myschemastore/main/schema1.json": {
+					id: 'https://myschemastore/schema1.json',
```
aeschli, line 326: Shouldn't that be `id: 'https://myschemastore/main/schema1.json'` ?

aeschli:
@mati-o Thanks a lot for the PR, it looks good!
I'm quite new to the feature but read https://json-schema.org/understanding-json-schema/structuring.html
- is `$anchor` the same as $id? Do we need to support it for refs?
- a guess is more work needed for [bundling](https://json-schema.org/understanding-json-schema/structuring.html#bundling)?

src/test/schema.test.ts
mati-o: Yes indeed it should be the same, although it doesn't affect the test, I would keep them the same for clarity. I'll fix it by the end of this day hopefully

mati-o:
> * is `$anchor` the same as $id? Do we need to support it for refs?

Apparently it is, never seen one using this though in my practice. Guess it could be addressed at some point later. 

> * a guess is more work needed for [bundling](https://json-schema.org/understanding-json-schema/structuring.html#bundling)?

I've never seen this form of schema but it indeed requires some effort. In our regular `$ref` case, we could split everything using the `#` character easily. However in this case it's trickier to know which `/` is part of the original URI and which isn't, without some relative URL code. I think it would be better to wait for someone to complain about it :)

aeschli:
I pushed two tests that show a problem when resolving $id across schemas.

The problem is when resolving schemas the `$id` nodes are removed. They can't be found anymore when the same schema is used a second time.
A solution would be to build up a map of id to node that we remember along with the resolved (or unresolved schema).

mati-o:
> I pushed two tests that show a problem when resolving $id across schemas.
> 
> 
> 
> The problem is when resolving schemas the `$id` nodes are removed. They can't be found anymore when the same schema is used a second time.
> 
> A solution would be to build up a map of id to node that we remember along with the resolved (or unresolved schema).
> 
> 

I'll give it a look towards the end of the week, sounds like an easy fix

src/services/jsonSchemaService.ts
```diff
@@ -527,17 +585,52 @@ export class JSONSchemaService implements IJSONSchemaService {
...
+				}
+			};
+
 			while (toWalk.length) {
 				const next = toWalk.pop()!;
 				if (seen.indexOf(next) >= 0) {
 					continue;
 				}
 				seen.push(next);
+				handleId(next);
```
mati-o, line 624: @aeschli I was thinking about the two tests you have added and since I already had a map from the full URI to the full schema, it got me thinking. The root cause of the problem was that the `handleRef` function was called before `handleId`, thus all the schemas with `$ref` and `$id` together were missing that `$ref`. By simply changing the order between them I have managed to solve these two cases without compromising the rest (luckily).

aeschli (changes requested):

aeschli:
I pushed a change to keep the map of resolved ids in the schema handle.

Sorry I had to 'force push' as merge always ended up with too many changes.

aeschli:
I think the current state is ready for testing and useful. so I'm pushing the PR.
We can test the new feature in the upcoming test week.

@mati-o Thanks for all your help!

Of course, there's still some gaps to fill. What comes to my mind:

- More tests, in particular from the JSON schema test suite: See #102
- Support for [anchor](https://json-schema.org/understanding-json-schema/structuring.html#id17)
- Support for [bundling](https://json-schema.org/understanding-json-schema/structuring.html#bundling)
- No need to resolve the full schema to find an anchor: https://github.com/microsoft/vscode-json-languageservice/blob/9d65fa79f0c2f8121dca783eeee0aee1c3e27857/src/services/jsonSchemaService.ts#L519

if you are interested in continuing to help, that would be fantastic! Let's continue in a new PR then.

aeschli (approved):

mati-o:
Thanks for merging! Glad I could help. I've been busy in the past two weeks so sorry for not being able to assist. I hope to find time in the near future to assist more

src/services/jsonSchemaService.ts
```diff
@@ -562,18 +569,16 @@ export class JSONSchemaService implements IJSONSchemaService {
...
 						// This is a reference inside the current schema
 						if (!seenRefs.has(ref)) {
-							if (isSubSchemaRef(segments[1])) { // A $ref to a sub-schema with an $id (i.e #hello)
-								// Get the full URI for the current schema to avoid matching schema1#hello and schema2#hello to the same
-								// reference by accident
-								const fullId = reconstructRefURI(parentSchemaURL, segments[1]);
-								tryMergeSubSchema(fullId, next);
+							const id = segments[1];
+							if (id !== undefined && isSubSchemaRef(id)) { // A $ref to a sub-schema with an $id (i.e #hello)
+								tryMergeSubSchema(next, id, handle);
```
michalchar, line 583: Pay attention you've replaced `parentSchemaURL` with `handle`. 
Therefore local refs in an external schemas wont work.

aeschli: Thanks, created #121

PR #111, opened 2021-10-19
interpret `pattern`/`patternProperties` regex as Unicode

jgraettinger:
This was under-specified in previous JSON schema drafts.

As of 2020-12, patterns are defined as Unicode:
https://json-schema.org/draft/2020-12/json-schema-core.html#regex

Issue #103


jgraettinger:
PTAL

KapitanOczywisty (approved):

Issue #116, opened 2021-11-12
Should state that JSON Schema 2020-12 is not supported to avoid false positives


Relequestual:
Any user who defines a JSON as using the 2020-12 dialect of JSON Schema will get a false positive that it is valid.
Example:
```
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "unevaluatedProperties": [],
  "items": []
}
```

In draft 2020-12, `unevaluatedProperties` is a new keyword, but it cannot be an array.
In draft 2020-12, `items` can no longer take an array (or tuple) form, so is invalid.
This JSON shows no errors. It would if 2020-12 was supported.

The reason for this I suspect is the implementation downloads the file located at the URI found in `$schema`. This is common for other use cases. However, the file located at that URI is a draft 2020-12 meta-schema, and is likely being treated as a regular draft-07 non-meta-schema. More on meta-schemas later.

**As an immediate fix, I propose calling out the lack of support for draft 2020-12.** (This was how lack of support for 2019-09 was handled previously.
As a later remediation, more work is required, as `$schema` may contain a network addressable URI which is a custom dialect using draft 2020-12 as it's basis (such as the OpenAPI dialect meta-schema). Doing so will currently cause no errors and a false positive on validation.

A meta-schema (using draft 2019-09 and up) can be identified by its use of the `$vocabulary` keyword, which is new from 2019-09. This new keyword defines the required vocabulary understanding to process a given JSON document (such as the one in the example). Only meta-schemas are allowed to `$vocabulary`.

For the official meta-schema, the `$schema` value and the `$id` value are the same, but this will not be the case for other dialects' meta-schemas. In those cases, you will need to continue to follow the `$schema` values until you find a known JSON Schema meta-schema. This determines which version (draft) of JSON Schema is being used, and therefore if you support it or not. This would be the later remediation.

As a further note, it's possible to bundle schemas which use different versions of JSON Schema into a single file.
Article: https://json-schema.org/blog/posts/bundling-json-schema-compound-documents
Implementation: https://github.com/hyperjump-io/json-schema-bundler

As such, ideally you would also check support for "embedded schema resources", which are identified by the use of `$id` when using draft 2019-09 and up. These embedded schema resources MAY declare their dialect support via the use of `$schema`, but usually they will not, as it will likely be the same as the enclosing resource and root document.

[...]

PR #117, opened 2021-11-12
Say JSON Schema draft 2020-12 is not yet supported

Relequestual:
Fixes https://github.com/microsoft/vscode-json-languageservice/issues/116


aeschli:
Makes sense, thanks @Relequestual !

GabenGar:
So which latest schema version is supported by VSCode then?

KapitanOczywisty:
@GabenGar `draft-07` is fully supported, anything above is partial or not at all.

PR #141, opened 2022-05-23
feat: support number const

bytemain:
we have a json schema:

```
      "version": {
          "type": "number",
          "const": 2,
          "markdownDescription": "version",
          "description": "version"
        },
```

we want its value can be auto suggestion


bytemain:
I will add a test later~

aeschli (approved):

lramos15 (approved):

aeschli:
Thanks @bytemain !

PR #148, opened 2022-07-14
New API `getLanguageService.sort` which sorts the JSONC files

aiday-mar:


src/jsonLanguageService.ts
```diff
@@ -95,17 +125,13 @@ export function getLanguageService(params: LanguageServiceParams): LanguageServi
...
-				const length = d.offsetAt(r.end) - offset;
-				range = { offset, length };
-			}
-			const options = { tabSize: o ? o.tabSize : 4, insertSpaces: o?.insertSpaces === true, insertFinalNewline: o?.insertFinalNewline === true, eol: '\n', keepLines : o?.keepLines === true };
-			return formatJSON(d.getText(), range, options).map(e => {
-				return TextEdit.replace(Range.create(d.positionAt(e.offset), d.positionAt(e.offset + e.length)), e.content);
-			});
-		}
+		sort: (d: TextDocument, o: FormattingOptions) => {
+			let originalJSONString = d.getText();
```
aeschli, line 129: I suggest to move that to a separate file, `sort.ts`

src/jsonLanguageService.ts
```diff
@@ -50,19 +50,49 @@ export interface LanguageService {
...
+	sort(document: TextDocument, options: FormattingOptions): string;
 	getFoldingRanges(document: TextDocument, context?: FoldingRangesContext): FoldingRange[];
 	getSelectionRanges(document: TextDocument, positions: Position[], doc: JSONDocument): SelectionRange[];
 	findDefinition(document: TextDocument, position: Position, doc: JSONDocument): Thenable<DefinitionLink[]>;
 	findLinks(document: TextDocument, doc: JSONDocument): Thenable<DocumentLink[]>;
 }
 
+function sortJSON(jsonObject: Record<any, any>): Record<any, any> {
+	let sortedJSON : Record<any, any> = {};
+	if (jsonObject.constructor.name === "Array") {
```
aeschli, line 62: Use `Array.isArray`, we avoid using the constructor name

src/jsonLanguageService.ts
```diff
@@ -50,19 +50,49 @@ export interface LanguageService {
...
+		sortedJSON = jsonObject;
+		sortedJSON.forEach(function (value : any, index : number) {
```
aeschli, line 64: Style: Use lambda (`(value, index) => { }`) 
As you assign to the same index you can also use `map`: `sortedJSON = jsonObject.map(v => sortJSON(v))` or even shorter:
`sortedJSON = jsonObject.map(sortJSON)`

src/jsonLanguageService.ts
```diff
@@ -50,19 +50,49 @@ export interface LanguageService {
...
+			sortedJSON[index] = sortJSON(value);
+		});
+	} else if (jsonObject.constructor.name === "Object") {
```
aeschli, line 67: we avoid using the constructor name use `val === 'object' && val !== null && !Array.isArray(val)` for which we already have a utility function in `objects.ts`: `isObject`

src/jsonLanguageService.ts
```diff
@@ -50,19 +50,49 @@ export interface LanguageService {
...
+		sortedJSON = {};
+		Object.keys(jsonObject).sort().forEach(function (key : any) {
```
aeschli, line 69: lambda is IMO nicer than function (and also allows you to use `this`, if you need to)

aeschli: avoid `any`. key is `string` here

src/test/sort.test.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    let options = { tabSize: 2, insertSpaces: true, keepLines : false, eol : '\n', insertFinalNewline : false};
+    
+    function sort(unsorted: string, expected: string, options: FormattingOptions) {
+        let document = TextDocument.create('test://test.json', 'json', 0, unsorted);
+        const sorted = ls.sort(document, options);
+        assert.equal(sorted, expected);
+    }
+
+	test.only('sorting simple JSON object without array', () => {
```
aeschli, line 17: remove the `only` (everywhere)

src/utils/sort.ts
```diff
@@ -0,0 +1,484 @@
...
+    propertyData["indexOfSecondDoubleQuoteInPropertyName"] = indexOfSecondDoubleQuoteInPropertyName;
+    return propertyData;
+}
+
+// When keepLines === true, the colon is not necessarily on the same line as the property name
+function findColonIndexAndLine(individualLines: string[], lineOfPropertyName: number, indexOfSecondDoubleQuoteInPropertyName: number) : Record<string, any> {
+    let colonData : Record<string, any> = {};
+    let colonIndex: number = individualLines[lineOfPropertyName].indexOf(':', indexOfSecondDoubleQuoteInPropertyName + 1);
+    while (colonIndex === -1) {
+        colonIndex = individualLines[lineOfPropertyName++].indexOf(':');
```
aeschli, line 160: lineOfPropertyName++ increments after reading -> index will one too big

aeschli (changes requested):

aeschli:
Instead of breaking the string up in characters ([...text]), use the scanner to tokenize the content. This will help with string literals and comment tokens that can also contain `{`, `,` `}`

PR #157, opened 2022-11-07
switch to l10n

aeschli:
For https://github.com/microsoft/vscode/issues/165729

### DIFF
--- a/src/services/configuration.ts
+++ b/src/services/configuration.ts
@@ -5,8 +5,7 @@
 
 import { ISchemaContributions } from './jsonSchemaService';
 
-import * as nls from 'vscode-nls';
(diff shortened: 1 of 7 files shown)


src/services/configuration.ts
```diff
@@ -526,7 +525,7 @@ for (const schemaName in schemaContributions.schemas) {
 		if (description) {
 			propertyObject['description'] = description;
 		} else {
-			console.log(`${property}: localize('schema.json.${property}', "")`);
+			console.log(`${property}: l10n.t("")`);
```
mjbvz, line 528: This looks like the right conversation but I don't think a `l10n` call is needed here since it's an empty string

mjbvz (approved):

PR #167, opened 2023-02-08
New API `sort` which sorts JSONC documents

aiday-mar:
This revised API uses the Scanner API to construct a PropertyTree which is used to sort the lines of the array. The sorted text is also formatted so that each property is on its own line using the `format` function. In the following description a property refers to an entity composed of a key and a value. Some notes about the PR:

- Each (line or block) comment is associated to the property below it, except for in three cases. 
1. If the comment is at the very end of an object or an array, then it is associated to the property above it. 
2. If the comment starts on the same line as the end of a property, then it is associated to it. An example could be for example:

```
"key" : "value", /**
...
**/
```
3. If a comment starts on the same line as an open brace, it is associated to it. An example could be:

```
{ /* ...
... */
}
```

- Upon sorting, the commas are added or removed as required.
- In the GIF below, you may find an example of the sorting of JSONC files.

https://user-images.githubusercontent.com/61460952/218491224-e8ec3c18-914c-4b7c-9c71-c929e7a6b221.mov


aiday-mar:
@microsoft-github-policy-service agree company="Microsoft"

src/utils/propertyTree.ts
```diff
@@ -0,0 +1,89 @@
...
+    commaLine : number | undefined;
+    lineWhereToAddComma : number | undefined;
+    indexWhereToAddComa : number | undefined;
+    type : Container | undefined;
+    childrenProperties: PropertyTree[];
+    lastProperty : boolean;
+    noKeyName : boolean;
+
+    constructor(
+        propertyName?: string,
```
aeschli, line 26: you can write `propertyName = ''`

src/utils/propertyTree.ts
```diff
@@ -0,0 +1,89 @@
...
+    childrenProperties: PropertyTree[];
+    lastProperty : boolean;
+    noKeyName : boolean;
+
+    constructor(
+        propertyName?: string, 
+        beginningLineNumber?: number
+        ) {
+
+        this.propertyName = propertyName ? propertyName : '';
```
aeschli, line 30: Note that `propertyName` is also true for empty string. It's better to write `propertyName !== undefined`. Or use the `??operator` which tests for undefined: `this.propertyName = propertyName ?? '';`

src/utils/sort.ts
```diff
@@ -0,0 +1,408 @@
...
+    const jsonPropertyTree : PropertyTree = findJSONPropertyTree(formattedJSONString);
+    const sortedArrayOfLines = sortLinesOfArray(arrayOfLines, jsonPropertyTree);
+    const sortedDocument: TextDocument = TextDocument.create('test://test.json', 'json', 0, sortedArrayOfLines.join('\n'));
+    const edits: TextEdit[] = format(sortedDocument, options, undefined);
+    return TextDocument.applyEdits(sortedDocument, edits);
+}
+
+function findJSONPropertyTree(formattedString : string) {
+
+    const arrayOfLines: string[] = formattedString.split('\n');
```
aeschli, line 31: It's better to avoid breaking up the file by lines and creating an array. Line delimiters can be `\n`, `\r` and '\r\n'.
Instead use the TextDocument to handle that.
The text document allows to access the content by line.

aeschli (changes requested):

aeschli:
There are some lint errors reported:
https://github.com/microsoft/vscode-json-languageservice/actions/runs/4175233855

aeschli (approved):

dbaeumer (approved):

aiday-mar:
For verification:

1. Open a JSONC document (with no formatting errors)
2. Add comments (block or line) and select 'JSON: Sort document' from the command palette. 
3. Check that the document is sorted as expected

PR #185, opened 2023-03-24
Insert missing comma or colon before completion

zardoy:
Inspired by https://github.com/microsoft/TypeScript/pull/52899
Fixes https://github.com/microsoft/vscode/issues/69545 (but I don't know what to do with *when `"` is typed)

Though when I just implemented it, I was thinking of why do we need "insert missing colon" behavior: if property has suggestions `:` will be added as insert text anyway (but not always e.g. doesn't work with `oneOf`). But if property doesn't have suggestions, there is no way to trigger such behavior. So I think we should make property completions to always include `:` as insert case (there are no shorthands in JSON 😃)


zardoy:
@aeschli I've removed colon insertion, so I do hope the pr will be easier to review :)

PR #187, opened 2023-04-06
Added function to register and clear external schema in the language service Interface

rahulbanerjee26:
This pull request adds a couple of functions to the Json Language Service Interface
- Function to register a schema
- Function to clear a previously registered schema.

Right now, if a user wants to register or clear schemas, the only way to do it is by using `configure`. Since `configure` overwrites previously registered schemas, this would require the user to maintain a list of schemas and update it accordingly and use it along with `configure`.

I noticed JSONSchema service has some utility functions that could make the above process easier. Please let me know your thoughts or if I misunderstood something and if there is currently a way to register/clean schemas without having to maintain a list of schemas.


rahulbanerjee26:

rahulbanerjee26:
@aeschli Could you please take a look when convenient?

PR #213, opened 2023-11-06
fix: unify doNotSuggest

p-spacek:
### What does this PR do?

related to this conversation https://github.com/redhat-developer/yaml-language-server/pull/911#issuecomment-1752466977

exclude doNotSuggest schema from autocompletion
`doNotSuggest` was already there but implemented only for the simple object properties.
This PR uses `doNotSuggest` also for other scenarios: anyOf, values

```json
"anyOf":[
  {
     "$ref": "schema1",
     "doNotSuggest": true
   },
  {
     "$ref": "schema2",
   }
]
```

## PROBLEM
this PR contains also a fix for the test utility `testCompletionsFor`, there is an issue with checking zero suggested items.
the previous implementation doesn't allow to check 'no result' with `expected: { count: 0}`
but when I fixed the test utils, the test `In comment` fails

Options: 
  - use this PR after the fix of the comments
  - if you give me advice on how to check empty results, I can modify this PR (but the comment issue remains)

### Is it tested? How?
adds unit tests


src/test/completion.test.ts
```diff
@@ -74,7 +74,7 @@ suite('JSON Completion', () => {
 		const position = Position.create(0, offset);
 		const jsonDoc = ls.parseJSONDocument(document);
 		return ls.doComplete(document, position, jsonDoc).then(list => {
-			if (expected.count) {
+			if (typeof expected.count === 'number') {
```
p-spacek, line 77: if you call this method with `expected.count === 0` it won't check zero length (`assert.equal(list!.items.length, expected.count`)

but after this change, it seems that test for `test('In comment', ...` is not correct

p-spacek:

aeschli (approved):

aeschli:
@p-spacek There's a test failure, if you can have a look...

roblourens (approved):

TylerLeonhardt (approved):

PR #219, opened 2024-01-14
Uri reference links

twelvelabs:
This PR adds logic to `findLinks` to create links for relative path names when the associated JSON schema is configured with `format: uri-reference`.

Fixes: microsoft/vscode#128971
Unblocks: redhat-developer/vscode-yaml#559

A few notes:

- I tried to make the change as backwards compatible as possible by adding a new, optional `schemaService` param to `findLinks`.

- This was important because the YAML language server [delegates to `findLinks`](https://github.com/redhat-developer/yaml-language-server/blob/dfccc6fc095faeb5d07051b51f308478cdac70fd/src/languageservice/services/yamlLinks.ts#L21), and I'm primarily interested in getting this functionality working over there.

- I assumed that the `fileExists` check would be a requirement (i.e. links for non existent files would be a poor UX). It's only being ran for properties w/ `format: uri-reference`, so it shouldn't be much overhead, but I'm happy to remove if you disagree.


twelvelabs:

src/services/jsonLinks.ts
```diff
@@ -6,24 +6,71 @@
...
+		if (node.type === "property" && node.valueNode?.type === 'string' && schemaService) {
+			const pathNode = node.valueNode;
+			const pathLinks: DocumentLink[] = [];
+			const promise = schemaService.getSchemaForResource(document.uri, doc).then((schema) => {
+				if (schema) {
+					const matchingSchemas = doc.getMatchingSchemas(schema.schema, pathNode.offset);
+					matchingSchemas.forEach((s) => {
+						if (s.node === pathNode && !s.inverted && s.schema) {
+							if (s.schema.format === 'uri-reference') {
+								const path = resolvePath(pathNode.value);
```
twelvelabs, line 51: Just occurred to me that this should probably be relative to the document URI - not the current working directory. Will update.

twelvelabs:
ping @aeschli

src/services/jsonLinks.ts
```diff
@@ -6,24 +6,75 @@
 import { DocumentLink } from 'vscode-languageserver-types';
 import { TextDocument, ASTNode, PropertyASTNode, Range, Thenable } from '../jsonLanguageTypes';
 import { JSONDocument } from '../parser/jsonParser';
+import { IJSONSchemaService } from './jsonSchemaService';
+import { URI } from 'vscode-uri';
+import { existsSync as fileExistsSync } from 'fs';
```
aeschli, line 11: We want to run that code also in the web, so we can not use NodeJS-API like `fs`.
Instead we need to provide file system access through an abstraction.
The already have that in the css-languageservice: https://github.com/microsoft/vscode-css-languageservice/blob/885c65cd38033a456225dd607a8d07776cd6bb6f/src/cssLanguageTypes.ts#L173

aeschli (changes requested):

src/services/jsonLinks.ts
twelvelabs: Sure! If I'm understanding things correctly, it looks like I'll need to use `stat` (I didn't see an exists function in the [fs provider docs](https://code.visualstudio.com/api/references/vscode-api#FileSystemProvider)) and handle the promise it returns. 

Then when I eventually want to use this in the yaml language server, I'm assuming I'll need to do something similar to [what you did here](https://github.com/redhat-developer/yaml-language-server/pull/539).

Please let me know if I'm not on right path.

aeschli: Looks like my link was corrupted, I updated it. 

What's to do:
- add new type to jsonLanguageTypes.ts:
```
export interface FileSystemProvider {
	stat(uri: DocumentUri): Promise<FileStat>;
}
```
plus `FileStat` and `FileType` (copy these from the [cssLanguageTypes](https://github.com/microsoft/vscode-css-languageservice/blob/3e2d7992179c81dc327c5e1ec8b2ec59cb2220f1/src/cssLanguageTypes.ts#L245C13-L245C21))

- add a new property 
```
	/**
	 * Abstract file system access away from the service.
	 * Used for dynamic link resolving, path completion, etc.
	 */
	fileSystemProvider?: FileSystemProvider;
``` 
to `LanguageServiceParams`
- pass that to your code and use it, if available (all is optional, we don't want to break existing clients)
- document these types in the CHANGELOG.md
- in followup PRs we have to update the consumers of the `vscode-json-languageservice` (yaml-extension and json-language-feature) to provide a FileSystemProvider implementation. I can do that for `vscode-json-languageservice`

twelvelabs: @aeschli Thanks for the detailed breakdown - made things much easier! I rebased against main and made the suggested changes.

I don't do much JavaScript development, and I'm not 100% happy w/ how complicated all the promise logic ended up where I'm checking if the file exists. If you have any suggestions for simplifying things, I'm all ears 😄.

src/jsonLanguageTypes.ts
```diff
@@ -252,6 +252,49 @@ export interface Thenable<R> {
...
+	 */
+	mtime: number;
+	/**
+	 * The size in bytes.
+	 */
+	size: number;
+}
+
+export interface FileSystemProvider {
+	stat(uri: DocumentUri): Promise<FileStat>;
```
aeschli, line 295: Please add specs here too, in particular what happens when the file does not exist.

aeschli (changes requested):

PR #222, opened 2024-02-21
Add `node` parameter to `JSONWorkerContribution.getInfoContribution`

### LINKED ISSUE #221: `getInfoContribution` is not composable
I'm struggling a little with the `JSONWorkerContribution` interface to provide smart hover.

From what I can read in [jsonHover](https://github.com/microsoft/vscode-json-languageservice/blob/main/src/services/jsonHover.ts#L56-L59), it looks like a `JSONWorkerContribution.getInfoContribution` should return undefined when there's nothing to show (which prevents the use of async function() since it would return Promise<undefined>).

I'm struggling with two things with this API design:
1. `getInfoContribution(uri, location)` does not let me discriminate on if the location is to a property or value node
2. If `getInfoContribution(uri, location)` returns something, hover info from the schema is not shown (it's one or the other)

I kind of would like to be able to show both if both have something. 

For instance, at Shopify we have those large translation files. It feels like it would make sense to have a getInfoContribution that prints out the translation key of the thing under the cursor (`'deep.path.to.key'`), but this should only happen on the key and if there's something special from the schema, then I'd probably want that as well. 

e.g. keys that end in `_html` have a special meaning and there's docs associated with them.

charlespwd:
This should be backward compatible and allow implementers to discriminate between value and key nodes in order to only optionally return `undefined | Promise<MarkedString[]>`

Fixes #221

### DIFF
--- a/src/jsonContributions.ts
+++ b/src/jsonContributions.ts
@@ -2,10 +2,10 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
-import { Thenable, MarkedString, CompletionItem } from './jsonLanguageService';
+import { Thenable, MarkedString, CompletionItem, ASTNode } from './jsonLanguageService';
 
 export interface JSONWorkerContribution {
-	getInfoContribution(uri: string, location: JSONPath): Thenable<MarkedString[]>;
+	getInfoContribution(uri: string, location: JSONPath, node: ASTNode): Thenable<MarkedString[]>;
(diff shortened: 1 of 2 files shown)


aeschli:
Adding `undefined` makes sense, but `node` would not help you discriminating property key or value at least not with the changes in the PR. `node` is always the value at that point. https://github.com/microsoft/vscode-json-languageservice/blob/bbba2d576f1983edee03c088a4d6bea70852d422/src/services/jsonHover.ts#L33

aeschli:
It should be `Thenable<MarkedString[] | undefined>`

aeschli:
Also if you can add a test case that would be great.

PR #223, opened 2024-02-29
Fix sorting error in case of nested trailing comma

denisw:
Fixes https://github.com/microsoft/vscode/issues/178229

When encountering a nested object property with a trailing comma, such as `{ "a": { "x": 1, "y": 2, }, "c": 3 }`, the `json/sort` command would throw the following error:

```
Range#create called with invalid arguments[[object Object], [object Object], undefined, undefined]: Error: Request json/sort failed with message: Range#create called with invalid arguments[[object Object], [object Object], undefined, undefined]
```

The reason is that the sorting code's scanner failed to navigate up to the parent when `}` after the trailing comma is encountered, because the condition `currentProperty!.endLineNumber === undefined` for doing so is false if the previous non-trivia non-comment token was a comma (`endLineNumber` would have already been set when processing the comma).

As a consequence, any further properties in the parent object (like `"c"` in the above example) would be wronly considered to be part of the preceding nested object, which confuses the scanner and results in the `endLineNumber` of the last property not being set at all. This, in the end, triggers the `Range.create()` error when the sorting code attempts build a range with that undefined end line number.

### DIFF
--- a/src/test/sort.test.ts
+++ b/src/test/sort.test.ts
@@ -53,7 +53,7 @@ suite('Sort JSON', () => {
             '"hobbies" : ["volleyball","drawing","hiking"],',
             '"friends" : {',
             '"Marc" : {"hobbies" : ["kayaking", "mountaineering"],',
-            '"age" : 35},',
+            '"age" : 35,},',
             '"Leila" : {"hobbies" : ["watching movies",',
             '"reading books"], "age" : 32}}}'
         ].join('\n');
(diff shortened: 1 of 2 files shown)


aiday-mar:
Thank you for the valuable contribution. I will have a look at the PR.

aiday-mar (approved):

aeschli (approved):

PR #237, opened 2024-07-19
Bugfix/43195 type completion issue

mpmartins:


mpmartins:

PR #238, opened 2024-07-21
Fix microsoft/vscode#209655: fix case-sensitive JSON sorting error

ttlopes:
**Summary**

This PR fixes the case-sensitive JSON sorting error described in microsoft/vscode#209655. The issue caused incorrect sorting of JSON properties with mixed-case names, like "test" and "Test".

**Details**

The problem was due to the sorting mechanism not handling case sensitivity properly, causing it to sort properties at random since it didn't know how to handle the mixed-case names. A new function was introduced to sort properties in a case-sensitive manner, ensuring accurate ordering of property names.

**Testing**

Two unit tests were added:

1. An unsorted array to verify it sorts correctly.
2. An already sorted array to confirm it remains unchanged.

These tests cover the description of the issue. 

This fix ensures that JSON properties are sorted correctly, addressing the reported issue.

### DIFF
--- a/src/test/sort.test.ts
+++ b/src/test/sort.test.ts
@@ -1435,4 +1435,52 @@ suite('Sort JSON', () => {
 
         testSort(content, expected, formattingOptions);
     });
+
+    test('sorting a JSON object with mixed case keys', () => {
+        const content = [
+            '{',
+            '  "tEst": "tEst",',
+            '  "tesT": "tesT",',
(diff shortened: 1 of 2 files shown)


ttlopes:

aeschli:
Thanks @ttlopes !

TylerLeonhardt (approved):

PR #239, opened 2024-07-30
Deprecate Thenable and alias to PromiseLike

remcohaszing:
TypeScript has a builtin type for promise-like values: `PromiseLike`. `Thenable` attempts to achieve the same result, but it has some compatibility issues.

This change aliases `Thenable` to `PromiseLike`, to make it compatible with the rest of the TypeScript ecosystem.

Additionally this deprecates `Thenable`. There is no practical use case for it, but it is confusing to communicate something other than `PromiseLike` to users. All occurrences of `Thenable` were replaced with `PromiseLike`.

### DIFF
--- a/src/example/sample.ts
+++ b/src/example/sample.ts
@@ -1,6 +1,6 @@
 
 
-import { getLanguageService, JSONSchema, SchemaRequestService, TextDocument, MatchingSchema } from '../jsonLanguageService';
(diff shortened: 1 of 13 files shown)


aeschli (approved):

aeschli:
Thanks @remcohaszing !

TylerLeonhardt (approved):

PR #243, opened 2024-09-21
feat(completion): support detail from schema

### LINKED ISSUE #235: [Feature Request] Json Schema show schema details. in CompletionItemLabel
As far as I know, the current [JSON worker](https://github.com/microsoft/vscode-json-languageservice/blob/main/src/services/jsonCompletion.ts) uses the `string` for the label property when handling completion items, rather than the `CompletionItemLabel` interface. Consequently, it cannot display inline brief hints in the suggestions. What I intend to convey is that I hope there could be a configuration option to control this behavior, allowing support for setting corresponding details on JSON schemas.
```typescript
export interface CompletionItem {
    label: string | CompletionItemLabel;
    // ...
}

export interface CompletionItemLabel {
    label: string;
    detail?: string;
    description?: string;
}
```
However, I would like to set the information in CompletionItemLabel#description or #detail to display it in the suggestions.
Because most of the time, I don’t want or need to hover over the document for this.

Sincerely looking for you reply.

sumimakito:
This pull request adds a new `completionDetail` field to `JSONSchema` to allow the customization of the `detail` field in the collected `JSONCompletionItem` results, which would be useful in scenarios like displaying type information along with the label of the completion item.

Closes #235

### DIFF
--- a/src/jsonSchema.ts
+++ b/src/jsonSchema.ts
@@ -85,6 +85,7 @@ export interface JSONSchema {
 	suggestSortText?: string;  // VSCode extension
 	allowComments?: boolean; // VSCode extension
 	allowTrailingCommas?: boolean; // VSCode extension
+	completionDetail?: string; // VSCode extension
 }
 
 export interface JSONSchemaMap {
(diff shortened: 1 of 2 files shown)


sumimakito:

aeschli:
Thanks @sumimakito !

aeschli (approved):

rzhao271 (approved):

lszomoru (approved):

PR #255, opened 2024-12-10
JSON sort: sort symbols first, then lower case, then upper case

aiday-mar:
JSON sort: sort symbols first, then lower case, then upper case

### DIFF
--- a/src/test/sort.test.ts
+++ b/src/test/sort.test.ts
@@ -1449,11 +1449,11 @@ suite('Sort JSON', () => {
 
         const expected = [
             '{',
-            '  "Test": "Test",',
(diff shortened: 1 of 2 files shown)


aeschli:
@aiday-mar I think the correct approach is to use [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare). Every locale has different sorting rules.

aiday-mar:
Hi @aeschli thanks for the review, I will change that right now

aiday-mar:
I made the change

aeschli (approved):

PR #261, opened 2025-04-16
feat: Notify Invalid RegExp Patterns

aedenmurray:
This PR allows users to be notified when an invalid regex pattern is encountered.
This is useful if a JSON schema is published that contains patterns that are not supported by `new RegExp()`.
For example, as of time of writing, the `version` pattern for the `composer.json` schema includes a possessive quantifier (`*+`) that is not supported by JavaScript (see [composer.json schema definition](https://github.com/composer/composer/blob/053a5ed2c7ba79533327554d6ac7cc8405ff3450/res/composer-schema.json)).

### DIFF
--- a/src/test/parser.test.ts
+++ b/src/test/parser.test.ts
@@ -614,12 +614,12 @@ suite('JSON Parser', () => {
 			properties: {
 				"one": {
 					type: 'string',
-					pattern: 'invalid pattern',
+					pattern: '*+ (invalid pattern)',
 				}
 			}
 		});
 
-		assert.strictEqual(semanticErrors!.length, 1);
+		assert.strictEqual(semanticErrors!.length, 0);
 
 		semanticErrors = validate2(jsonDoc, textDoc, {
 			type: 'object',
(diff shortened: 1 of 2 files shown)


aedenmurray:
@microsoft-github-policy-service agree company="Shadowscape, Inc."

src/test/parser.test.ts
```diff
@@ -609,6 +609,18 @@ suite('JSON Parser', () => {
...
+			type: 'object',
+			properties: {
+				"one": {
+					type: 'string',
+					pattern: 'invalid pattern',
+				}
+			}
+		});
+
+		assert.strictEqual(semanticErrors!.length, 1);
```
aedenmurray, line 622: I'm not entirely sold on the need for this test (or at least, how it functions).
Let me know what you think.

aedenmurray:
Thanks @aeschli!
You are correct - this is an issue with `composer.json`.
However, the current implementation attempts to blame the user, which is misleading:
Screenshot 2025-04-25 at 7 55 29 PM
I agree with the sentiment that invalid patterns should be ignored.
I have updated the PR to match.

aeschli (approved):

alexr00 (approved):

PR #268, opened 2025-05-19
use https locations for all old schema drafts

aeschli:


jeremyfiel:
@aeschli  This pr introduced invalid JSON Schema `$schema` values

All of the old drafts ONLY use `http` because the identifier is `http://json-schema.org/draft-0(4/7)/schema#`

this was originally fixed in #201  seems to be a different location where you made these changes.  I just found intellisense is again suggesting the wrong uris. 

image

Issue #263, opened 2025-04-25
Let contributors suggest array item completions


leonrinkel:
Currently, triggering completion inside an array does not collect completion items from contributors. It would be nice to be able to suggest additional completion items beyond the "New object" filler item, especially when the array is supposed to contain strings rather than objects.

PR #276, opened 2025-06-20
feat: array value completion from contributions

leonrinkel:
This PR enables `JSONWorkerContribution` to contribute array value completions. I originally requested this feature in #263 and have now taken a first stab at implementing it. To avoid changing the API, I reused `collectValueCompletions`, but it doesn't feel like the best fit for this use case. Another approach might be introducing something like `collectArrayCompletions`, possibly with better fitting attributes like the index the cursor is at. I'm happy to hear your thoughts and feedback!

### DIFF
--- a/src/services/jsonCompletion.ts
+++ b/src/services/jsonCompletion.ts
@@ -525,6 +525,30 @@ export class JSONCompletion {
 				}
 			});
 		} else {
+			if (
+				node.parent?.type === 'array' && node.parent?.parent?.type === 'property' &&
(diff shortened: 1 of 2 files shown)


aeschli:
Thanks for looking at this!
I agree that it would be better to add `collectArrayCompletions` as an new (optional) callback.
It might look like that:
`collectArrayCompletions(uri: string, location: JSONPath, index: number, result: CompletionsCollector): PromiseLike<any>`

PR #283, opened 2025-08-22
Escape plain text description hover correctly

Legend-Master:
This matches the behavior with the documentation shown in the autocomplete

With this schema:

```json
{
  "properties": {
    "test": {
      "description": "For example:\n\n<script>\n  alert(1)\n</script>\n\n    Test [1]"
    }
  }
}
```

Before:

image

After:

image

### DIFF
--- a/src/services/jsonHover.ts
+++ b/src/services/jsonHover.ts
@@ -60,61 +60,64 @@ export class JSONHover {
 		}
 
 		return this.schemaService.getSchemaForResource(document.uri, doc).then((schema) => {
-			if (schema && node) {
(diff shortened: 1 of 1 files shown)


aeschli:
The reformatting makes it hard to discover what the fix is. Can you explain?

aeschli (dismissed):

Legend-Master:
> Test failure...

Should be fixed now

leia-uwu:
why are spaces being replaced with non line breaking spaces (`&nbsp;`) ?

this makes text wrapping on long tooltips worse
on vscode for example it will break in the middle of words and even insert spaces as the first character of new lines, and in some other editors like kate it will prevent wrapping at all and add a horizontal scrollbar

PR #292, opened 2025-10-26
feat: add discriminator support for JSON schema validation

mohankumarelec:
### Overview
This PR introduces discriminator support for JSON schema validation, significantly improving performance for schemas using `oneOf`, `anyOf`, and `allOf` by intelligently selecting which schema alternatives to test based on a discriminator property.

### Problem Solved

Without discriminators, the validator must test **all** schema alternatives in `oneOf`/`anyOf`/`allOf`, leading to exponential complexity in deeply nested or self-referencing schemas. This causes severe performance degradation and can even make validation impractically slow for certain schema patterns.

See the test case `self-referencing schema with anyOf and deep nesting (exploding complexity test)` for a concrete example.

### Implementation

Added discriminator support with two modes:

1. **`propertyName`**: Discriminate based on an object property (e.g., `"type": "cat"`)
2. **`propertyIndex`**: Discriminate based on the first N properties or array items

The discriminator can include an explicit `mapping` object to map discriminator values to schema indices, or auto-detect based on `const` values in the alternative schemas.

**Example:**
```typescript
{
  oneOf: [
    { properties: { type: { const: 'cat' }, meow: { type: 'boolean' } } },
    { properties: { type: { const: 'dog' }, bark: { type: 'boolean' } } }
  ],
  discriminator: {
    propertyName: 'type',
    mapping: { 'cat': 0, 'dog': 1 }
  }
}
```

### Test Coverage

Added tests for:
- ✅ Basic discriminator with `propertyName` and `propertyIndex`
- ✅ Discriminator with `oneOf`, `anyOf`
- ✅ Missing/invalid discriminator values (graceful fallback)
- ✅ Complex nested objects
- ✅ Non-string discriminator values
- ✅ Auto-detection without explicit mapping
- ✅ **Deep nesting with self-referencing schemas (100 levels)**

### Performance Impact

For schemas with discriminators, validation performance improves significantly. This is especially critical for self-referencing schemas with deep nesting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 5.6.0 / 2025-05-28
 ================
- * added `Schema.enumSortTexts`and  `Schema.enumDetails` to control the sort order and presentation of suggestions for enaums
+ * added `Schema.enumSortTexts` and `Schema.enumDetails` to control the sort order and presentation of suggestions for enums
 
 5.5.0 / 2025-03-25
 ================
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "vscode-json-languageservice",
-  "version": "5.6.2",
+  "version": "5.6.4",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "vscode-json-languageservice",
-      "version": "5.6.2",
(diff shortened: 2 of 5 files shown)


mohankumarelec:
@aeschli , Can you please look into this PR when you have some time. Thanks a lot

PR #325, opened 2026-05-23
feat: expose default value in hover

rchl:
I don't know if you are interested in such a feature and it's fine if not but I thought I'll try to upstream it instead of just having a local patch.

Screenshot 2026-05-23 at 14 43 22

Screenshot 2026-05-23 at 14 46 32


sirreal:
`examples` has the same gap:

```json
{ "a": { "type": "string", "description": "D", "default": "dflt", "examples": ["ex1"] } }
```

Hovering `"a"` gives `["D"]`. Both annotations are dropped, even though completion at the value position already proposes them (`"dflt"` with `detail: "Default value"`, `"ex1"` with none).

PR #337, opened 2026-07-07
feat: support $dynamicRef / $dynamicAnchor (JSON Schema 2020-12) 

keegan-caruso:
The simple case: A `$dynamicAnchor` is registered as a normal anchor, the same way `$anchor` is, so references can find it. A `$dynamicRef` resolves to wherever it points lexically. For many schemas that's all you need, and it behaves like a plain `$ref`. This is the first commit.

The second part is dynamic resolution. To resolve dynamically, you need to know which anchors belong to which schema resource. The problem is that `$ref` merging flattens those resource boundaries before we get a chance to look. So, we run a pre-pass (`collectDynamicAnchors`) over the original tree first. It records, for each resource, which anchors it has (the dynamic ones and all of them), and which resource each `$dynamicRef` was written in. A `$dynamicRef` resolves its first target inside its own resource, so an anchor with the same name in a sibling resource can't be picked up by mistake.

At validation time the rule is this: if the first target is a `$dynamicAnchor` with the right name, we look through the current dynamic scope from the outside in and use the outermost match. If it isn't, the reference stays where it is and acts like a plain `$ref`. In both cases the `$dynamicRef` is applied together with any sibling keywords on the same node, not in place of them.

That pre-pass has a limit worth calling out: it runs over the document being validated. Within a single document the dynamic scope is solid, and that's most of what you'll hit. The weak spot is bookending that has to reach into a *separate* document. A `$dynamicAnchor` that overrides a base default can live in another document reached by `$ref`, and there it never joined the scope, so the override was lost. The reference still found its starting point the normal way — it just didn't extend the scope across the document boundary the way the root does. The fix is to collect a referenced document's anchor maps too, once, while the external link is resolved (again, before merging flattens the boundaries), and fold them in with the same outermost-wins rule. Each resource's draft is read from its own `$schema`, so pulling in a pre-2020-12 document doesn't hand it `$dynamicAnchor` semantics.

Updating the vendored test suite to current main turned up a handful of related reference-scope bugs, in 2019-09 as much as 2020-12. None are in `$dynamicRef` itself — they're places where `$ref` or `$recursiveRef` merging was too eager or quietly dropped something:

• A `$recursiveRef` stopped after validating its target, so a sibling `unevaluatedItems` or `unevaluatedProperties` on the same node never got a turn. It now falls through, and they see the annotations the reference produced.
• When the referenced and referencing schemas, both constrain array items, a flatten-merge could only keep one of them. Those cases isolate into an `allOf` so both apply — and the anchor and `unevaluated*` keywords stay on the wrapper, so bookending targets stay findable and `unevaluated*` still sees every applicator's results.
• A boolean subschema — `true` or `false` — is a real schema, not a missing section, when a `$ref` points at one.
• A referenced document's own embedded `$id` resources are registered as targets, so a nested `$ref` by that id resolves to them instead of being fetched again as a fresh document.
• Only the dynamic anchor map is folded across resources; the per-resource map stays put, so a sibling's identically named anchor can't leak into a `$dynamicRef`'s lexical lookup.

Two smaller points:

• The internal data (the per-resource anchor maps, and each ref's first target and scope) is stored in non-enumerable fields so it doesn't show up in serialized schemas. It's kept in two records, `DynamicRefInfo` on the reference and `AnchorMaps` on the resource root, rather than several separate properties.
• In drafts older than 2020-12, `$dynamicRef` is treated as an unknown keyword and ignored, so nothing changes for existing schemas.

Testing

New project-owned tests cover the scenarios above — the `$dynamicRef`/`$dynamicAnchor` edge cases, the cross-document overrides, and the reference-scope fixes — so none of these leans entirely on the vendored suite. The vendored [JSON-Schema-Test-Suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite) is bumped to current main (`92acb61`) and runs across draft4/6/7/2019-09/2020-12 with nothing skipped: 5326 tests, all passing.

Code coverage

Line: 97.5%
Branch: 93%
Func: 95%

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+6.0.0-next.2 / 2026-07-11
+================
+- Support $dynamicRef / $dynamicAnchor (JSON Schema 2020-12)
+- Support JSON Schema 2019-09
+
 6.0.0-next.1 / 2026-03-04
 ================
 * Breaking: the package is now ESM-only (`"type": "module"`) and exposes ESM entry points through `exports`.
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
         "@typescript-eslint/eslint-plugin": "^8.57.0",
         "@typescript-eslint/parser": "^8.57.0",
         "eslint": "^9.39.4",
-        "json-schema-test-suite": "https://github.com/json-schema-org/JSON-Schema-Test-Suite.git#69acf52990b004240839ae19b4bec8fb01d50876",
+        "json-schema-test-suite": "https://github.com/json-schema-org/JSON-Schema-Test-Suite.git#92acb61eb772a932c077d5ffa634ded719d2d738",
         "rimraf": "^6.1.3",
         "typescript": "^5.9.3"
       }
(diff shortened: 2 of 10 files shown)


keegan-caruso:
FYI @jdesrosiers and @aeschli

baywet:
Cc @handrews and @karenetheridge

aeschli (dismissed):

lszomoru (dismissed):

TylerLeonhardt (dismissed):

vijayupadya (approved):

vijayupadya (dismissed):

chrmarti (approved):

aeschli (approved):

PR #342, opened 2026-07-15
feat: adds support for schema store

baywet:
this pull request adds support for json schema store so schemas are suggested automatically based on https://www.schemastore.org/ when no $schema is present


aeschli:
Schema store loading and management should not be part of the JSON language service.
- we don't want to hard-code any schema registry URL, nor it's file format
- the requestService (type SchemaRequestService) is spec to load schemas, not catalogs. It's API and implemented by users of the service. VS Code is only one of them
- there are already APIs to pass schema associations via the LanguageSettings. I would first try to get it working with these. If that doesn't work, add a (optional) property `LanguageServiceParams.readFaillbackSchema: (resource: string) => string | undefined`

baywet:
@aeschli thanks for the feedback. I've moved the code that resolves the catalog, etc... to the extension itself.
Let me know if you have any additional comments or questions.

baywet:
superseded by https://github.com/microsoft/vscode/pull/327628

PR #343, opened 2026-07-17
Fix: select matching deprecated alternative in anyOf/oneOf (#304)

### LINKED ISSUE #304: Deprecated matching string pattern inside `anyOf` is not considered a match
The following schemas should produce the same output on the `test.json` file:

`test1.schema.json`:
```json
{
    "$schema": "https://json-schema.org/draft/2020-12/schema",
    "type": "object",
    "properties": {
        "key": {
            "anyOf": [
                {
                    "type": "string",
                    "const": "unused literal"
                },
                {
                    "type": "string",
                    "const": "text value",
                    "deprecated": true
                }
            ]
        }
    }
}
```

`test2.schema.json`:
```json
{
    "$schema": "https://json-schema.org/draft/2020-12/schema",
    "type": "object",
    "properties": {
        "key": {
            "anyOf": [
                {
                    "type": "string",
                    "const": "unused literal"
                },
                {
                    "type": "string",
                    "pattern": "^text value$",
                    "deprecated": true
                }
            ]
        }
    }
}
```

`test.json`:
```json
{
    "key": "text value"
}
```

However, using `test2.schema.json` shows `Value is not accepted. Valid values: "unused literal".` on hover instead of the expected `Value is deprecated` that it shows for `test1.schema.json`. 

Image

This means that the text `"text value"` did not match the pattern `"^text value$"`, which it should... (and does when `"deprecated": true` is removed)

Note that this example seems to be minimal: removing the non-deprecated branch of the `anyOf` leads to different (and correct) results, for example

ATKasem:
Fixes #304.

In an `anyOf`/`oneOf`, a branch that actually matches the value but is marked `deprecated` was treated as a non-match, so a non-matching sibling got reported instead. With the schema from the issue, `"text value"` matches the `pattern` branch, but the hover/validation showed `Value is not accepted. Valid values: "unused literal".` instead of `Value is deprecated`.

The cause is that the deprecation notice is pushed as a problem, and `ValidationResult.hasProblems()` counted it like a validation error when selecting the best alternative. The `const` case in the issue only worked by luck — `enumValueMatch` is a separate tiebreaker in `compare()`; a `pattern` branch has nothing equivalent, so the deprecated-but-matching branch lost.

Fix: exclude deprecation problems from `hasProblems()` (they're advisory, `severity: Warning`, `code: ErrorCode.Deprecated`). Match selection now ignores them, while the deprecation diagnostic itself is still reported.

Added a test in `parser.test.ts` covering the issue's schema (matching value → `Deprecated`; non-matching value → the non-match error, unchanged). Full `parser.test.js` passes; `jsonSchemaTestSuite` failure count is unchanged from `main` on my machine (pre-existing URN cases).


ATKasem:
@aeschli — gentle ping, open since July 17.

Small fix (+36/-1): in an `anyOf`/`oneOf`, a branch that matches the value but is marked `deprecated` is treated as a non-match, so a non-matching sibling gets reported instead. With the schema from #304, `"text value"` reports `Value is not accepted. Valid values: "unused literal"` rather than `Value is deprecated`. The cause is `hasProblems()` counting the deprecation notice like a validation error when selecting the best alternative.

Let me know if you'd prefer the deprecation notice tracked separately from problems rather than filtered at the comparison site — happy to restructure.

aeschli (approved):

aeschli:
Thanks @ATKasem !

dmitrivMS (approved):

roblourens (approved):

PR #345, opened 2026-07-24
fix: id related lookups for rooting and path segmentation

### LINKED ISSUE #344: Base URI is determined incorrectly for local file schemas
According to the JSON Schema specification (https://json-schema.org/draft/2020-12/json-schema-core#section-9.1.1-3), and RFC3986 (https://www.rfc-editor.org/info/rfc3986#section-5.1.1), if a schema specifies an `$id`, the Base URI for relative references should come from the `$id`, and not from the Retrieval URI.

However, this is not the case in vscode. Consider the following project layout:

**first.schema.json**
```json
{
    "$id": "http://example.com/first",
    "$schema": "https://json-schema.org/draft/2020-12/schema",
    "type": "object",
    "properties": {
        "test": {
            "$ref": "second#$defs/somestruct"
        }
    }
}
```

**second.schema.json**
```json
{
    "$id": "http://example.com/second",
    "$schema": "https://json-schema.org/draft/2020-12/schema",
    "$defs": {
        "somestruct": {
            "type": "object",
            "properties": {
                "innerTest": {
                    "type": "string"
                }
            }
        }
    }
}
```

**.vscode/settings.json**
```json
{
    "json.schemas": [
        {
            "url": "http://example.com/first",
            "schema": {
                "$ref": "file:///home/vince/commandpod/WrongURIBaseRepro/first.schema.json"
            }
        },
        {
            "url": "http://example.com/second",
            "schema": {
                "$ref": "file:///home/vince/commandpod/WrongURIBaseRepro/second.schema.json"
            }
        }
    ]
}
```

**example.json**
```
{
    "$schema": "http://example.com/first"
}
```

`example.json` produces the following warning:

> Problems loading reference '$defs/somestruct': Unable to load schema from '/home/vince/commandpod/WrongURIBaseRepro/second': Schema not found: file:///home/vince/commandpod/WrongURIBaseRepro/second.

Which indicated that vscode is interpreting the reference in `first.schema.json` relative to its Retrieval URI, `file:///home/vince/commandpod/WrongURIBaseRepro/first.schema.json`, rather than its `$id`, `http://example.com/first`. (Which should ideally result in `http://example.com/second`, and resolve to `second.schema.json` through the mapping in `settings.json`.)

baywet:
fixes #344

### DIFF
--- a/src/services/jsonSchemaService.ts
+++ b/src/services/jsonSchemaService.ts
@@ -613,7 +613,7 @@ export class JSONSchemaService implements IJSONSchemaService {
 			return handle.anchors.get(id);
 		};
 
-		const getSchemaId = (schema: JSONSchema): string | undefined => schema.$id || schema.id;
+		const getSchemaId = (schema: JSONSchema): string | undefined => schema.$id || schema.id || (schema as MergedJSONSchema).$originalId;
 
 		// Fold a source resource's $dynamicAnchor names into a target node's dynamic
 		// map, creating it on demand. Existing entries win, so the outermost resource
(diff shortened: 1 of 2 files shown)


aeschli:
replaced by https://github.com/microsoft/vscode-json-languageservice/pull/359

Issue #143, opened 2022-06-22
Improve handling of schema properties having default values


Delapouite:
Hello

## Summary

Many JSON schema have a handy `default` value for properties. Currently it is used by the auto-completion feature which is nice!
I suggest that it could also be used for other situations:

- in the `doHover`. The tooltip appearing on hover of key currently displays the property description (in Markdown). It would be great to also display the `default` value in this tooltip.
- if a validated JSON has a key which happens to have the default value as its current value, in this case it could be consider a redundant info and a developer may be interested by removing this line from the JSON. I suggest this could potentially be surfaced as a `warning` or something similar.

## Concrete example

Let me illustrate with a concrete situation, involving opening a `.prettierrc.json` file. This is a config file used by the popular Prettier code formatter. When this kind of file is opened in VSCode, a dedicated extension correctly set https://json.schemastore.org/prettierrc as the schema.

If the content of this `.prettierrc.json` is :
```
{
  "endOfLine": "lf"
}
```

- When I hover `endOfLine` with my mouse, it displays `Line Feed only (\\n), common on Linux and macOS as well as inside git repos` in the tooltip, which is the `description` of this enum. That's great. But it would be event better if this tooltip would additionally displays something like ` (default : "lf")`
- In this scenario, as `"lf"` is the default value of `endOfLine`, it means that this config is "useless" because it only repeat the default behavior of this end of line option. As a developer I would find it valuable if I was able to have some kind of warning telling me just that : each case where I needlessly respecify a key, with no added benefit.

Thanks a lot

michaeltlombardi:
Having this information in the hover view would also be useful for context when I'm reviewing existing data - is the current value the default? What _is_ the default for this property, if any?

WilliamK112:
I would like to take the hover portion of this issue as a focused PR. I will append the schema's JSON-serialized default value to property/value hover content while preserving title, Markdown description, and enum-value descriptions, with regression coverage for string, `0`, `false`, `null`, and structured defaults.

I plan to leave the proposed "redundant explicit default" diagnostic out of this change: explicitly pinning a default can be intentional and a new warning policy deserves separate design discussion.

PR #362, opened 2026-09-09
Show schema default values in hover

WilliamK112:
## Summary
- Append a schema's JSON-serialized default value to property and value hover content, labeled **Schema default** so consumers such as VS Code Settings do not confuse it with an effective value inherited through configuration scopes.
- Preserve existing title, Markdown description, and enum-value description rendering.
- Handle falsey defaults such as `0`, `false`, and `null`, along with strings and structured JSON values.

## Scope
This addresses the hover portion of #143. It intentionally does not add a diagnostic for explicitly configured default values, since pinning a default can be intentional and that warning policy needs separate design.

## Validation
- Node 22.14.0: `npm test` (5,343 tests passed; lint passed)
- Follow-up `ff790e2`: TypeScript compile, focused hover regression, ESLint, and `git diff --check`
- `git diff --check`

## AI assistance
Codex assisted with implementation and test execution. I reviewed the final diff and understand the behavior and tests.

### DIFF
--- a/src/services/jsonHover.ts
+++ b/src/services/jsonHover.ts
@@ -67,11 +67,15 @@ export class JSONHover {
 			let title: string | undefined = undefined;
 			let markdownDescription: string | undefined = undefined;
 			let markdownEnumValueDescription: string | undefined = undefined, enumValue: string | undefined = undefined;
+			let defaultValue: string | undefined = undefined;
 
 			const matchingSchemas = doc.getMatchingSchemas(schema.schema, node.offset).filter((s) => s.node === node && !s.inverted).map((s) => s.schema);
 			for (const schema of matchingSchemas) {
 				title = title || schema.title;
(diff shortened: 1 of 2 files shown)


WilliamK112:
Thanks, that is fair feedback. I agree that “Schema default” exposes an implementation detail to Settings users, and adding an nls dependency solely for that label would be hard to justify.

The original #143 example uses a third-party schema, so putting the value in its existing description is not something the JSON document author can do. But I agree that automatically appending raw schema defaults is a product decision, not just a hover implementation detail. An optional `defaultMarkdownDescription` would make the intent explicit for schema authors, although it would not change hovers for existing schemas unless they opt in.

Would you prefer that I pivot this PR to the opt-in `defaultMarkdownDescription` behavior you suggested, with a focused regression test, or close it for now? I will hold off on further automatic-default changes until we agree on the direction.

aeschli:
If the optional solutions are not helpful, then I think we should leave things as they are.

PR #375, opened 2026-09-17
Validate calendar dates in date and date-time formats

mikevillari:
The `date` and `date-time` formats currently accept impossible calendar dates, including `2023-02-29` and `2023-04-31T12:00:00Z`. Their regular expressions allow days through 31 without checking the month or leap year.

Add a shared calendar check after the existing syntax check, following [RFC 3339 section 5.7](https://www.rfc-editor.org/rfc/rfc3339.html#section-5.7). Regression tests cover month boundaries, leap-year century rules, early years, timestamp offsets, and diagnostic messages and ranges.

Validation:
- `npm test` on Node 22.23.2: compilation, all 5,352 tests, and all 9 lint checks pass. Both new regression tests fail before the fix.
- All 69 upstream JSON Schema optional date-format cases pass through the public API.
- A separate calendar-boundary check passed 48,096 comparisons, including a full 400-year Gregorian cycle.

AI-assisted with Codex.


mikevillari:

